Issue #2, opened 2015-07-05
Create a version of `handleActions` which supports async actions too


johanneslumpe:
We want to be able to support handlers in the form of:

```javascript
{
    begin: (state, action) => {},
    end: {
      next: (state, action) => {},
      throw: (state, action) => {},
    }
  }
```

acdlite:
I have a branch https://github.com/acdlite/redux-actions/tree/sequence that solves this pretty well. I'll update the docs soon and get it merged in.

kpaxqin:
@johanneslumpe @acdlite 
Here is [my idea](https://github.com/kpaxqin/redux-action-tools#createreducer) 

Example here: 

```js
const handler = (state, action) => newState

const reducer = createReducer()
  .when([ACTION_FOO, ACTION_BAR], handlerForBothActions) // share handler for multi actions
  .when('BAZ', handler)    // optimistic update here if you need
  .done(handler)               // handle 'BAZ_COMPLETED'
  .failed(errorHandler)      // handle 'BAZ_FAILED'
  .build(initValue); 
```

dthree:
I don't know why async isn't even mentioned in the docs. Making at least one async call in any real web app seems unavoidable. Am I missing something? [Currently looking at this library](https://github.com/machadogj/redux-thunk-actions) to handle it.

PR #7, opened 2015-07-07
Async actions

johanneslumpe:
This allows us to handle async actions, see #2

### DIFF
--- a/src/__tests__/handleActions-test.js
+++ b/src/__tests__/handleActions-test.js
@@ -9,7 +9,22 @@ describe('handleActions', () => {
 
       DECREMENT: ({ counter }, { payload: amount }) => ({
         counter: counter - amount
-      })
+      }),
+
+      ASYNC_TEST: {
+        begin: ({ counter}, { payload: amount }) => ({
+          counter: counter + amount * 2
+        }),
+        end: {
+          next: ({ counter}, { payload: amount }) => ({
+            counter: counter + amount * 3
+          }),
+
+          throw: ({ counter}, { payload: amount }) => ({
+            counter: counter + amount * 4
+          })
+        }
+      }
     });
 
     expect(reducer({ counter: 3 }, { type: 'INCREMENT', payload: 7 }))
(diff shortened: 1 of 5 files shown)


johanneslumpe:
@acdlite please review and check if i missed anything :)

acdlite:
This looks good. Do you mind if we wait a bit before merging though? Still not totally sure this approach is best. It'd be nice to find a solution that works for observables, too (e.g. polling a server).

johanneslumpe:
Sure, let's wait a bit and see if we come up with something better. Polling a server though would always dispatch an action with the 'end' status, right? It just wouldn't have `error` set, so it would use the `next` function.

ralphschindler:
Has a best practice emerged for this yet? Thanks.

marr:
Any updates on this?

laurentdesmet:
Is this something that will be merged somewhere in the future? Async actions are fairly common so it would be nice if there is some kind of standard solution for this using the redux-actions library instead of implementing a custom async implementation using the `meta` field.

timche:
We can't merge this PR, but we will definitely work on that asap. If we the author is willing to update this PR, we can make it happen here.

alexFaunt:
@timche @acdlite  Any thoughts on this since it was last considered?

I've been thinking of:
```javascript
{
first() {
  // If payload is a promise this is fired
},
next() {
 // if payload is an object, but not an error this is fired
},
throw() {
 // if payload is an error object this is fired
}
}
```

This allows first class handling of pending actions without a breaking change. I've got a working version as a fork and happy to clean it up for a PR if you see fit. 

But if you were to use something more like a stream for polling (mentioned above), I'd suggest renaming the handlers to next, done and throw. Where 'next' processes any streamed values and 'done' being called with the final value. Promises would call 'next' once at the start, and then 'done' when complete. This would be a big breaking change though, and would need more thought and proper use cases setting out as i'm not sure theres a huge demand for such behaviour.

Issue #3, opened 2015-07-06
how to set meta property?


ustccjw:
As title, how to set meta property?

clearjs:
@ustccjw Could you please elaborate on the problem and context that you have? `meta` is just a field in a plain action object, its "properties" are fields as well.

ustccjw:
return (...args) => ({
        type,
        payload: finalActionCreator(...args)
    })

I mean, when use `createAction`, the return value is payload, i want to pass meta property, how can i do?

acdlite:
@clearjs I think @ustccjw wants to be able to specify meta using `createAction()`. I've been thinking about it, because it's particularly important for async actions, where you need the ability to pass certain metadata to the `throw()` handler, and to the `begin()` handler once we implement https://github.com/acdlite/redux-actions/issues/2

@johanneslumpe and I discussed this on Slack, and the idea we come up with is to have an optional second parameter to `createAction()` that accepts the same parameters as the action creator, but it returns a plain object whose contents will be added to the `meta` field:

```js
const addTodo = createAction('ADD_TODO', (text, id) => ({ text, id }), (text, id) => ({ id }));
addTodo('Use Redux', 123); // { type: 'ADD_TODO', payload: { text: 'Use Redux', id: 123 }, meta: { id: 123 } }
```

acdlite:
That way, for example, if the action fails (say you're using redux-promise and the promise rejected) the reducer still has access to the meta object, which it can use to undo an optimistic update:

```js
{ type: 'ADD_TODO', error: true, payload: error, meta: { id: 123 } }
```

clearjs:
@ustccjw @acdlite Thanks. Sorry, I was still thinking in the context of FSA and thus didn't get it at first.

acdlite:
Oops, forgot to close this, but v0.8 added support for a meta creator https://github.com/acdlite/redux-actions/releases/tag/v0.7.0

PR #8, opened 2015-07-08
Allow to fill `meta` prop of an action

johanneslumpe:
This addresses #3 and will be useful together with #2.

@acdlite Does it seem logical for you to just use `identity` if no meta creator has been passed? Seems better to me than omitting `meta` in that case.


acdlite:
I'm thinking we should probably omit it. The action creator may receive something that's not serializable, like a promise. It's easy enough to pass an identity function for those special cases.

johanneslumpe:
Ok. I'll change this to omit if it no function is passed asap.

acdlite:
Thanks @johanneslumpe!

istarkov:
Thank you all. 
@acdlite `npm publish`, please.

PR #16, opened 2015-08-15
implement FSA errors by setting error to true if payload is an Error obj

ngokevin:
Noticed this spec of FSA wasn't yet implemented, and there's an issue for it.

This sets action.```error``` to true if the payload is a direct instance of ```Error```. This is the *faintest* touch of magic, but if it's documented, it makes sense.

Also fixes a possible error where we're trying to set action.meta even though action is a ```const```.

r? @acdlite

### DIFF
--- a/src/createAction.js
+++ b/src/createAction.js
@@ -13,7 +13,14 @@ export default function createAction(type, actionCreator, metaCreator) {
       payload: finalActionCreator(...args)
     };
 
-    if (typeof metaCreator === 'function') action.meta = metaCreator(...args);
+    if (args.length === 1 && args[0] instanceof Error) {
+      // Handle FSA errors where the payload is an Error object. Set error.
+      action.error = true;
+    }
+
+    if (typeof metaCreator === 'function') {
+      action.meta = metaCreator(...args);
+    }
 
     return action;
   };
--- a/README.md
+++ b/README.md
@@ -27,6 +27,23 @@ expect(increment(42)).to.deep.equal({
 });
 ```
 
+If the payload is an instance of an [Error
+object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Error),
+redux-actions will automatically set ```action.error``` to true.
+
+Example:
+
+```js
+increment = createAction('INCREMENT');
+
+var error = new TypeError('not a number');
+expect(increment(error)).to.deep.equal({
+  type: 'INCREMENT',
+  payload: error,
+  error: true
+});
+```
+
 **NOTE:** The more correct name for this function is probably `createActionCreator()`, but that seems a bit redundant.
 
 Use the identity form to create one-off actions:
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -51,5 +51,24 @@ describe('createAction()', () => {
         }
       });
     });
+
+    it('sets error to true if payload is an Error object', () => {
+      const actionCreator = createAction(type);
+      const errObj = new TypeError('this is an error');
+
+      const errAction = actionCreator(errObj);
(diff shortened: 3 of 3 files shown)


acdlite:
Not sure about this... Feels like it's outside the scope of the module. if the payload is an error, that means someone caught the error and placed it there. Shouldn't it be up to them to set `error` to `true`?

ngokevin:
But there doesn't seem to be a way to set ```error``` to ```true``` with ```createAction```. The only thing that can be set by the user is ```type``` and ```payload```.

acdlite:
Oh duh I see what you mean. I was thinking about `handleAction()` for some reason — actually I dunno what I was thinking.

This seems reasonable then.

src/createAction.js
```diff
@@ -8,12 +8,19 @@ export default function createAction(type, actionCreator, metaCreator) {
...
 
   return (...args) => {
-    const action = {
+    let action = {
       type,
       payload: finalActionCreator(...args)
     };
 
-    if (typeof metaCreator === 'function') action.meta = metaCreator(...args);
+    if (args.length === 1 && args[0].constructor === Error) {
```
acdlite: This check won't work for subclasses of Error (e.g. TypeError)

ngokevin: Ah thanks, updated to use ```instanceof```.

ngokevin:
Updated to use instanceof

ricardofbarros:
+1

krasnoperov:
+1

cappslock:
+1

tkqubo:
+1

jonny-improbable:
+1, please can you consider releasing this, @acdlite?

bravewick:
+1

pbomb:
I'm running into this issue now as well. Is this issue expected to be resolved soon?

tokenvolt:
+1

danilotorrisi:
I've also had the same issue, I hope it's gonna be released soon!

src/createAction.js
```diff
@@ -8,12 +8,19 @@ export default function createAction(type, actionCreator, metaCreator) {
     : identity;
 
   return (...args) => {
-    const action = {
+    let action = {
```
lukewestby: Any reason to change this from `const` to `let`?

ngokevin: i guess i didn't have to. thought maybe i should since we modify action below, but we don't reassign.

lukewestby: To me it's less about the semantics of the keyword and more about consistency with the rest of the code base. `const` is used whenever a variable does not need to be reassigned.

lukewestby:
@ngokevin this is awesome, thanks for submitting! Can you update the README to document your change?

Does this change look like it will be a breaking change? I can't think of any scenarios where this would cause existing code not to work but if anyone can then we will need @acdlite's input before merging.

ngokevin:
@lukewestby thanks for checking it out, updated.

Strate:
:+1:

kpaxqin:
@ngokevin 
seems we can't customize the error object by payloadCreator. 
maybe we can do in this way

```js
export default function createAction(type, actionCreator, metaCreator) {
    ? actionCreator
    : identity;

  return (...args) => {
    const payload = finalActionCreator(...args)
    const action = {
      type,
      payload,
      error: (payload instanceof Error)
    };

    if (typeof metaCreator === 'function') {
      action.meta = metaCreator(...args);
    }

    return action;
  };
}

```

acdlite:
Huzzah! I finally got off my ass and merged this! Thanks!

kpaxqin:
@acdlite 
I've create a PR to set error to true when payloadCreator return Error object.

PR #17, opened 2015-08-17
Refactored default parameters of createAction function to avoid …

PSpSynedra:
…typechecks and potential undefined behavior when passing arbitrary objects instead of functions.

Hi acdlite, thanks for the nice project redux-actions. We'd like to propose a change of default parameter handling of the factory function createAction. This avoids undefined behavior of createAction when passing non function parameters as metaCreator or actionCreator.

Furthermore it ensures, that the meta field of the resulting action object is always defined (unless of course the metaCreator explicitly returns undefined ;-)

### DIFF
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -25,7 +25,8 @@ describe('createAction()', () => {
       const action = actionCreator(foobar);
       expect(action).to.deep.equal({
         type,
-        payload: foobar
+        payload: foobar,
+        meta: {}
       });
     });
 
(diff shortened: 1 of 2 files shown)


lukewestby:
Thanks for your contribution, @PSpSynedra!

+1 for using `identify` as a default arg. i'm not sure about always including `meta` though, as it isn't required to be present by FSA and middleware may exist that rely in its presence or absence for their behavior.

PSpSynedra:
Ok. I understand.

Just read the flux-standard-action readme again and there it is clearly stated, that is MAY have a meta field. Thus my change pull request does not really make sense ;-)

PR #19, opened 2015-08-21
Sequence - use naming closer to the observable spec.

tappleby:
This PR is not necessarily intended to be merged but brings the sequence naming closer to the observable spec + the preferred naming discussed in acdlite/flux-standard-action#7.

Ive also tweaked the reducer to pass any extra arguments to the handler, this allows for cases when you need to pass dependant state into the handlers.

This version is published as a scoped package if you want to try it out: [@tappleby/redux-actions](https://www.npmjs.com/package/@tappleby/redux-actions)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,12 +43,12 @@ Wraps a reducer so that only handles Flux Standard Actions of a certain type.
 
 If a single reducer is passed, it is used to handle both normal actions and failed actions. (A failed action is analogous to a rejected promise.) You can use this form if you know a certain type of action will never fail, like the increment example above.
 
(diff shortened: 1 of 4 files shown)


acdlite:
Thank you for the PR. I'll merge the method renaming, but I don't agree with the need for passing extra arguments to the reducer. If you want to inject extra stuff to a reducer, you can use a higher-order function:

```js
function myReducer(dependentState) {
  return handleActions({
    [SOME_ACTION_TYPE]: (state, action) => {
      // Use state, action, and dependent state here
    }
  }, initialState)
}
```

tappleby:
The wrapped reducer makes much more sense.

I'm going to clean this up a bit and re-submit.

PR #28, opened 2015-09-06
showcase it with an import statement

capaj:


lukewestby:
Thanks for your contribution @capaj! I 100% agree with the need for this, but ultimately ended up adding a single demonstration of how to import all functions at the top of the README.

PR #33, opened 2015-09-25
Ensure that the global Reflect object also has an ownKeys function

dbachrach:
We are using traceur and it generates a Reflect global without `ownKeys`. This commit further tests the Reflect global to ensure it has the needed function.


yoavniran:
yes please. Using jspm with babel and handleActions fails with "_Reflect.ownKeys is not a function_"

dbachrach:
Thanks for merging! Are you planning on making a 0.8.1 release soon?

PR #34, opened 2015-10-01
Update eslint and config, add flowtypes

clentfort:
Add flowtypes and update eslint and babel-eslint to more recent versions. Switch
the default eslintrc from airbnb's to facebook's.

I took the liberty to update to the `.eslintrc` and fixed any linting errors that now were reported.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,7 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[options]
--- a/package.json
+++ b/package.json
@@ -26,10 +26,11 @@
   "devDependencies": {
     "babel": "^5.6.14",
     "babel-core": "^5.6.15",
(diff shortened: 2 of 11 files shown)


clentfort:
Another side-effect of this PR: If a non-function is passed either in a `ReducerMap` context or directly as a `Reducer` `handleAction` will try to use this nevertheless and ultimately this will cause an error. The previous approach of just returning the current state makes it very hard to debug because there is no real error if I pass in the wrong type, my state just wont change. In situations where I as a developer pass in a non-function I would expect the entire thing to blow up instead of doing nothing. Blowing up makes it easier to find errors.

clentfort:
Closing this PR so I can recreate if from another branch an free up my master-branch. Sorry about the noise.

PR #35, opened 2015-10-04
Adjust `handleAction` to new ES6 observers interface, fixes #30

### LINKED ISSUE #30: Match updated Observer interface
https://github.com/zenparsing/es-observable#observer

```js
interface Observer {

    // Recieves the next value in the sequence
    next(value);

    // Recieves the sequence error
    error(errorValue);

    // Receives the sequence completion value
    complete(completeValue);
}
```

`handleAction()` / `handleActions()` should use the same naming scheme:

```js
const reducer = handleAction(ACITON_TYPE, {
  // Receives normal actions
  next(state, action);

  // Receives `error: true` actions
  error(state, action);

  // Receives `meta.sequence: 'complete'` actions
  complete(state, action);
});
```

clentfort:
This PR changes the shape of `ReducerMap` that `handleAction` accepts. It now acccepts `error` instead of `throw` to handle errors, furthermore it now also knows about `first` and `complete` and will use either one depending on the `meta.sequence`-property of the action.

This PR is based on #36 .

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,7 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[options]
--- a/README.md
+++ b/README.md
@@ -43,12 +43,14 @@ Wraps a reducer so that only handles Flux Standard Actions of a certain type.
 
 If a single reducer is passed, it is used to handle both normal actions and failed actions. (A failed action is analogous to a rejected promise.) You can use this form if you know a certain type of action will never fail, like the increment example above.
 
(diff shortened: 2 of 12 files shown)


alexgorbatchev:
This PR also appears to be adding Flow typing.

Also, if you are suggesting API change, please include updated README and examples of use cases to open up a proper discussion.

clentfort:
The text clearly states it is based on pr #36, which adds FlowTypes. If you check the code carefully you will see that it updates the README according to the changes, tests are included which can guide as examples. If you would also take a look at the date of this PR, it was created over 6 month ago and has gone stale since then. I don't really recall what issues were open at that time but I assure you it made sense at the time.

PR #37, opened 2015-10-05
Keep flowtypes during compliation

clentfort:
Keep the flowtypes introduced in #36 during the compilation, making them available to consumers of redux-actions.

### DIFF
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -1,4 +1,5 @@
 import { createAction } from '../';
+import identity from '../identity';
 import isPlainObject from 'lodash.isplainobject';
 
 describe('createAction()', () => {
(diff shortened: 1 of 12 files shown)


acdlite:
Hey thanks for all your PRs. I will get these merged soon... my attention has been elsewhere lately. I appreciate your efforts. :)

clentfort:
No worries, actually I'm not sure this one should be merged yet. I didn't have the time to check if this even works propperly with flow. There are some issues open at the babel-plugins/babel-plugin-flow-comments. Maybe it makes more sense to ship a `lib` of pure flow types.

src/__tests__/createAction-test.js
```diff
@@ -1,31 +1,32 @@
...
 
     it('has no extraneous keys', () => {
-      const actionCreator = createAction(type, b => b);
+      const actionCreator = createAction(type, identity);
       const foobar = { foo: 'bar' };
       const action = actionCreator(foobar);
       expect(action).to.deep.equal({
         type,
-        payload: foobar
+        payload: foobar,
```
alexgorbatchev: I suggest making syntactical changes as a separate PR. Here it just introduces noise and makes it harder for reviewers to grok the actual change.

alexgorbatchev:
I think for introducing typing, TypeScript is a better target because it has significantly larger adaptation.

clentfort:
How does this contribute to this PR? Flow and TypeScript are not mutually exclusive, so what is your point?

alexgorbatchev:
I don't think the source should be converted to Facebook Flow.

clentfort:
And neither should it be converted to TypeScript. You could have stated your opinion directly instead of posting some trollish comments.

If you had payed a bit more attention to this PR, you would have noticed it was opened in October 2015 and since than has gone stale.

rcsole:
Hi @clentfort nice effort,

Is there any chance the work in these PRs could be distilled in a contribution
to https://github.com/marudor/flowInterfaces?

faceyspacey:
@clentfort I think @alexgorbatchev meant the source rather than library definitions. That said I agree with @clentfort that having one definition has no effect on the other, but this PR has changes to the source.

That said, has anyone seen or built (perhaps just partially) a flow definition file for redux-actions? It's not so fun to switch to flow and have your type snow *flow* through redux-actions :(

faceyspacey:
@clentfort I created an external flow type definition for `createAction` to start, but am I'm running into some issues in getting real use out of it. I discuss it in this issue:

https://github.com/acdlite/redux-actions/issues/183

I was wondering if you had any thoughts on the matter??

PR #40, opened 2015-10-29
Fix issue when rejecting a promise for a single reducer

andreisoare:
Fix for the following issue:

If I'm using handleAction with a single reducer rather than a reducerMap, then rejecting a promise will not make it to the reducer.


lukewestby:
Since the stated behavior in the README is "If a single reducer is passed, it is used to handle both normal actions and failed actions." this is a bug and this fix should be merged in. I can do it later today if no one else wants to.

lukewestby:
@acdlite thanks for merging! Just realized today that this could cause existing code to break if it depends on the prior behavior.

PR #43, opened 2015-11-19
Sets error to true when payloadCreator return an Error object

kpaxqin:


lukewestby:
To me it makes more sense to check the resulting payload for an error than to check that the arg passed in was an error. +1 from me.

danny-andrews:
I personally think that if you pass an error in as the payload, it should bypass your custom `payloadCreator` logic and just set the payload equal to the passed-in error. Otherwise, you have to add an error check in each of your action creators to achieve the same effect.
Example:
```javascript
const createCoolAction = createAction(COOL_ACTION_TYPE, spec => {
  const {zipCode, id} = spec;
  return {body: {zipCode, id}};
});
```
So, if I do something like, `createCoolAction(new Error('Something bad happened'))`, the payload will be set to `{body: {zipCode: error, id: undefined}}` instead of just the error object itself.

timche:
@yangmillstheory I think we can close this one because it is already [in master](https://github.com/acdlite/redux-actions/blob/master/src/createAction.js#L20-L23), am I right?

yangmillstheory:
I think this is a bit different; what I see in `master` is https://github.com/acdlite/redux-actions/blob/master/src/createAction.js#L9.

This is saying that the `payloadCreator` is allowed to return an `Error`, but we bypass the `payloadCreator` entirely in `master` and just check the first action creator argument.

yangmillstheory:
Closing this as stale. Also, this duplicates the newer https://github.com/acdlite/redux-actions/pull/43.

PR #48, opened 2015-12-16
Make payload optional (as per FSA spec)

john-mortlock:


ipanasenko:
I +1 this

danny-andrews:
+1. Just ran into a problem with this. Although, I think checking `args.length === 1` instead of `payload !== undefined` would be more robust, because what if someone wanted to explicitly set their payload to undefined for some reason (odd to be sure, but we should allow it).

timche:
@john-mortlock I am currently just checking this PR, because I was added to this repo and I would like to have this feature as well.

@danny-andrews made a good point for checking `args.length === 1`.

We have a few more cases were don't want to have `payload`. I currently have an action that looks like `createAction('TYPE', null, () => ({foo: bar})()`. As you can see, the `payload` is `null` and only `meta` should be set as an object `meta: {foo: bar}`. With this PR, the payload would still be `null`.

I only have that example above, because I can't think of another one, but I think there are one or two more. Any chances you could cover some more cases (like this)?

I would be happy to merge this feature in.

alexgorbatchev:
@timche It's not super clear to me what your request is. Are you asking for more test coverage?

timche:
@alexgorbatchev `payload` shouldn't only be optional when it is `undefined`.

It's more common to write

```js
let actionCreator = createAction('FOO', null, (baz) => ({bar: baz}))
```

than

```js
let actionCreator = createAction('FOO', undefined, (baz) => ({bar: baz}))
```

My request is to check `payload` also for `null`.

But it doesn't look that @john-mortlock is still active on GitHub. I will go probably for an own PR.

john-mortlock:
Hi, I think I have updated the pull request.. I am new to Git and Github so all a bit foreign to me.

timche:
Hey, your PR has still some conflicts, can you resolve them please? Otherwise I can help you with that if you want.

john-mortlock:
Pretty sure its merged in now.

timche:
Thanks. Just had a look at it again and I think now `payload` can't be set when it's also `0` or `''`. Correct me if I'm wrong, but can you write tests for them as well to check it?

A better approach would be to do `if (!(payload === null || payload === undefined)`. With that way, we have more control.

john-mortlock:
You right those values couldn't be set, I have adjusted as per your code and added some more tests to cover those values.

src/createAction.js
```diff
@@ -9,10 +9,14 @@ export default function createAction(type, actionCreator, metaCreator) {
 
   return (...args) => {
     const action = {
-      type,
-      payload: finalActionCreator(...args)
+      type
     };
 
+    const payload = finalActionCreator(...args);
+    if (!(payload === null || payload === undefined)) {
```
timche: Sorry, but it can be actually simplified with `if (payload != null)`. This will also return `false` for `null` or `undefined` only.

timche: Nvm, this would break the eslint rule `eqeqeq`.

acdlite:
👍

timche:
Thanks @john-mortlock, merged :)

PR #32, opened 2015-09-25
Add support for action creators used as object keys

rstuven:
```javascript
    it('can be used as object key', () => {
      const actionCreator1 = createAction('TYPE1');
      const actionCreator2 = createAction('TYPE2');
      const handlers = {
        [actionCreator1]: () => {},
        [actionCreator2]: () => {}
      };
      expect(Object.keys(handlers)).to.have.length(2);
    });
```

PR #51, opened 2015-12-26
Use action function as handleAction(s) type param

nikita-graf:
This PR allows to use actions as type parameter for `handleAction` and `handleActions` like so:
```javaScript
let increment = createAction('INCREMENT', amount => amount);

  [increment]: (state, action) => ({
    counter: state.counter + action.payload
  })
}, { counter: 0 });

handleAction(increment , {
  next(state, action) {...}
  throw(state, action) {...}
});
```
It prevents from copy&paste action names across modules, allows to use autocomplete in IDE.


src/createAction.js
```diff
@@ -6,8 +6,7 @@ export default function createAction(type, actionCreator, metaCreator) {
   const finalActionCreator = typeof actionCreator === 'function'
     ? actionCreator
     : identity;
-
-  return (...args) => {
+  let actionHandler = (...args) => {
```
julien-f: IMHO, it should be `const` instead of `let`.

src/handleAction.js
```diff
@@ -5,9 +5,13 @@ function isFunction(val) {
 }
 
 export default function handleAction(type, reducers) {
+  const typeValue = isFunction(type)
```
julien-f: Simply use `String(type)`.

nikita-graf: Do you mean `const typeValue = String(type)` instead of `const typeValue = isFunction(type) ? type.toString() : type` ?

julien-f: Yes, the `String()` function converts the passed value to a string, calling the `toString()` method if necessary.

nikita-graf: We can't always convert `type` to string type because it could be a `Symbol`. There is a test for that https://github.com/acdlite/redux-actions/blob/master/src/__tests__/handleActions-test.js#L25 . By the way `toString()` has a slightly better performance https://jsperf.com/string-vs-tostring2

julien-f: Those are excellent reason, sorry for bothering you :D

nikita-graf: Thanks for review! :)

timche:
Currently duplicated with #32, but when the author does not answer there, we can go with this PR. Are you fine with that?

nikita-graf:
Yes, I'm ok with that. I just want to have this feature merged through this PR or [#32](https://github.com/acdlite/redux-actions/pull/32), it doesn't matter.

timche:
Looks like the author of #32 doesn't care anymore of his PR, so we'll go with this one.

src/createAction.js
timche: Yup, please use `const` here.

timche:
Can you also provide some tests please?

nikita-graf:
@timche I've added 2 tests and replaced `let` with `const`. Please review it.

timche:
@nikita-graf Thanks, but `This branch has conflicts that must be resolved` ;)

nikita-graf:
Didn't see that, for now it's up to date:smile:

timche:
Thanks @nikita-graf, merged :)

nikita-graf:
Great, thanks  :)

PR #52, opened 2015-12-28
Did not you think that using action.type in handler callback is redundant? 

coderany:


lukewestby:
Actions may also have a `meta` property attached to them which needs to be available in the reducer. If we want to make this kind of change, the reducers will need to be passed the `meta` value as an optional third argument. It makes sense to me, but @acdlite will need to weigh in. I would also ask that with code change PRs you please make sure the related unit tests pass, it looks like there are 6 failing tests with this change. Thanks!

PR #63, opened 2016-02-07
only create reducer function one time

contra:
creating a new function w/ a spread each time is a little wasteful


acdlite:
Hmm what do you mean by "each time"? The function is only called once.

acdlite:
`handleActions()` creates a function (a reducer) which is called many times, but `handleActions()` itself is only ever called once.

acdlite:
Oh duh nevermind. This makes sense. Thanks!

PR #64, opened 2016-02-15
support adding meta to error actions

jreinert:
this allows creating actions like this:

``` js
const actionCreator = createAction('TYPE', null, (_, meta) => meta);
actionCreator(new Error('foobar'), { foo: 'bar' });
// => { type: 'TYPE', payload: Error: 'foobar', meta: { foo: 'bar' }, error: true }
```

### DIFF
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -67,5 +67,18 @@ describe('createAction()', () => {
       });
       expect(isFSA(noErrAction)).to.be.true;
     });
+
+    it('sets error to true if payload is an Error object and meta is provided', () => {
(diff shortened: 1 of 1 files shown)


jreinert:
please note that travis isn't failing because of any failing tests

timche:
Can you please rebase? I will have a look at it again then, but looks good so far.

jreinert:
It's a bit obsolete now since there was some refactoring to the action creator apparently. The test is maybe nice to have anyway. It's up to you, of course, but I won't cry tears of blood if you decide not to merge ;)

timche:
Oh, yeah, forgot about that change. I'm fine with that test, I'll merge it, thanks :)

Issue #65, opened 2016-02-15
Set payload equal to error if payload is `instanceof Error`


danny-andrews:
I already made a note of this in #43, but I wanted to promote it to its own issue.

I personally think that if you pass an error in as the payload, it should bypass your custom payloadCreator logic and just set the payload equal to the passed-in error. Otherwise, you have to add an error check in each of your action creators to achieve the same effect.
Example:
```js
const createCoolAction = createAction(COOL_ACTION_TYPE, spec => {
  const {zipCode, id} = spec;
  return {body: {zipCode, id}};
});
```
So, if I do something like, `createCoolAction(new Error('Something bad happened'))`, the payload will be set to `{body: {zipCode: undefined, id: undefined}}` instead of just the error object itself. To fix this, I could add an error check, but I don't want to have to do this in every one of my action creators.

I'd be happy to make an MR for this.

andrewsuzuki:
I agree. Just made a PR, see #66

yangmillstheory:
Wasn't this closed by https://github.com/acdlite/redux-actions/commit/b9d5eb7daec5e29cb02c34dfb855c51183051f71, and a duplicate of https://github.com/acdlite/redux-actions/issues/82?

PR #66, opened 2016-02-16
Bypass action and meta creators if payload is an Error object

andrewsuzuki:
See #65.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,7 @@
     "prepublish": "make clean build"
   },
   "files": [
-    "src",
-    "lib"
+    "lib/*.js"
   ],
   "keywords": [
     "flux",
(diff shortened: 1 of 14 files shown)


luisincrespo:
+1

timche:
Imo it shouldn't bypass the meta creators, unless there is a good reason to do so. If the meta creator is needed to handle the action, the action would maybe become useless without the meta information with this PR.

What do you think?

package.json
```diff
@@ -34,6 +34,7 @@
     "chai": "^3.0.0",
     "eslint": "^0.24.0",
     "eslint-config-airbnb": "0.0.6",
+    "escope": "3.3.0",
```
yangmillstheory: Is this being used anywhere?

timche: Nope, should be removed.

andrewsuzuki: Not sure if I'm remembering correctly, but I think I needed to add that to bypass an issue with babel-eslint back when redux-actions was on babel 5. I'll remove it now.
https://github.com/estools/escope/issues/99

andrewsuzuki:
@timche Just saw this. I agree, there could definitely be meta information associated with an error.

andrewsuzuki:
@timche Done, meta creator will still be run regardless of error now. Also my bad, should have rebased instead of merged.

yangmillstheory:
Is this still merge-able?

timche:
Sorry for the late reply, I was on vacation.

src/createAction.js
```diff
@@ -6,17 +6,20 @@ export default function createAction(type, actionCreator, metaCreator) {
   const finalActionCreator = typeof actionCreator === 'function'
     ? actionCreator
     : identity;
+
   const actionHandler = (...args) => {
+    const hasError = args[0] instanceof Error;
```
timche: Can you use `if (action.payload instanceof Error)` [here](https://github.com/acdlite/redux-actions/blob/master/src/createAction.js#L19-L22) instead of `args[0]` please?

timche:
Merged, thanks for the changes :)

kujon:
It totally reverted https://github.com/acdlite/redux-actions/pull/75

PR #76, opened 2016-03-29
Use flux-standard-action dependency for test only

### LINKED ISSUE #60: flux-standard-action dependency not necessary
handleAction use **flux-standard-action** method **isError**

> import { isError } from 'flux-standard-action';

the method is simple, but it will import some lodash method when compile by webpack.

```
function isError(action) {
  return action.error === true;
}
```

I suggest remove the flux dep for descrease the compilse file size.

albertogasparin:
Including FSA means adding unneeded dependencies. We keep using it inside tests so we ensure consistency. Closes #60 
Note: I've removed "has no extraneous key" test because `isFSA()` already does that check.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,10 +38,10 @@
     "eslint-config-airbnb-base": "^1.0.3",
     "eslint-plugin-import": "^1.5.0",
     "lodash.isplainobject": "^3.2.0",
+    "flux-standard-action": "^0.6.0",
     "mocha": "^2.2.5"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


timche:
Would be happy to merge this but tests are failing and we have some merge conflicts, because I've fixed the master build. Could you fix this PR?

albertogasparin:
Hi @timche, branch updated and looks like it's all green now 👍

timche:
Thanks, merged :)

PR #78, opened 2016-04-04
Fix build

jimf:
Conservatively bump babel-eslint to fix upstream bug that was breaking this build, as well as fix a missing semi-colon eslint violation.


timche:
Fixed with https://github.com/acdlite/redux-actions/commit/feaedd398668f00963cfde846b2f676aae109378.

PR #79, opened 2016-04-16
Add type property to generated action creator

jupl:
This comes in handy to have a singular representation of an action, whether using the function for dispatching or the type for reducers/sagas/etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ let increment = createAction('INCREMENT', amount => amount);
 // same as
 increment = createAction('INCREMENT');
 
+expect(increment.type).to.equal('INCREMENT')
 expect(increment(42)).to.deep.equal({
   type: 'INCREMENT',
   payload: 42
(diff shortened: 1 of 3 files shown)


timche:
Duplicated with #32, but imo not as simple as #32. I will close this PR, sorry.

jupl:
While #32 is a nice solution, it doesn't cover something like redux-saga's `take`/`takeEvery`/`takeLatest`/etc.

timche:
Will this PR cover these things? I haven't used `redux-saga` yet, sorry. I more detailed explanation would be helpful for me to understand how this PR is different to #32.

jupl:
The basic idea is to have access to the action type string accessible. In the case of `redux-saga` actions can be intercepted by the `type` string. http://yelouafi.github.io/redux-saga/docs/api/index.html#takepattern

Originally I had this as a constants file but especially with #32 it's becoming very redundant to have a file with type constants and another file with actions declared with `redux-actions` when basically the latter can be fully replaced without the use of an extra constants file.

PR #62, opened 2016-02-06
upgrade eslint-related things

lukewestby:
To incorporate [this bug fix](https://github.com/babel/babel-eslint/issues/243) that caused CI failures after merging just README changes

cc @acdlite

PR #81, opened 2016-04-22
Upgrade ESLint dependencies and fix tests

timche:
- Upgrade `eslint` to latest version
- Remove `babel-eslint`, because `eslint` has ES6 support
- Remove `eslint-config-airbnb`
- Add and use `eslint-config-airbnb-base`
- Add `eslint-plugin-import`
- Overwrite `eslint-config-airbnb-base` rules to match current style (except `arrow-body-style`)
- Fix missing `semicolon`

This would close #78 and #62.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,5 @@
 {
-  "extends": "eslint-config-airbnb",
+  "extends": "airbnb-base",
   "env": {
     "mocha": true,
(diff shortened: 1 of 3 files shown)


timche:
@acdlite Are you fine with that? :) I'd also like to upgrade to Babel 6 after this.

PR #83, opened 2016-04-27
Introduce new function: createActions

alanrsoares:
Allow the user to create a map of standard actions from a actionType map

example:

file: constants/actionTypes.js
```js
export const MY_FIRST_ACTION_TYPE = 'MY_FIRST_ACTION_TYPE'
export const MY_SECOND_ACTION_TYPE = 'MY_SECOND_ACTION_TYPE'
```

file: actions/index.js
```js
import { createActions } from 'redux-actions'
import * as actionTypes from '../constants/actionTypes'

const actionCreators = createActions(actionTypes)
// { myFirstActionType: createAction('MY_FIRST_ACTION_TYPE'),  
//   mySecondActionType: createAction('MY_SECOND_ACTION_TYPE'),  }

export default actionCreators
```

cc @acdlite


alexgorbatchev:
👍 

IMO, this PR should include updated README as well to be considered.

timche:
I actually can't say if this useful or not. We could need a bit more opinions on that.

jeffbski:
I was thinking that it would nice to have a createActions that could take either an array or an object and they might work like this. 

Makes it easy to create a bunch of identity actions and/or actions with custom PayloadCreators.

```js
// if called with an array then create actions for each and 
// assign to map by key
const mapActions = createActions(['foo', 'bar', 'baz']); // performs the following code internally
/*
const mapActions = {
  foo: createAction('foo'),
  bar: createAction('bar'),
  baz: createAction('baz')
};
*/

// if called with an object then use the key for the type and 
// the value becomes the PayloadCreatorFn if it is a fn
const mapActions = createActions({
  foo(text) {
    return text; // payload
  },
  bar(id, text) {
    return { id, text }; // payload
  }
});
// performs the following code internally
/*
const mapActions = {
  foo: createAction('foo', text => text),
  bar: createAction('bar', (id, text) => { id, text })
};
*/

// we could also allow null value to default to PayloadIdentityFn
// this example uses Identity for foo, then provides PayloadCreatorFn for bar
const mapActions = createActions({
  foo: null, // use default payload identity fn
  bar(id, text) {
    return { id, text }; // payload
  }
});
// performs the following code internally
/*
const mapActions = {
  foo: createAction('foo', text => text),
  bar: createAction('bar', (id, text) => { id, text })
};
*/
```

felixmc:
Here's my own external module I use in some personal projects that shares a lot of functionality with this PR: https://www.npmjs.com/package/redux-actions-magic

It uses the following syntax to generate both types and action creators from an array of string types:

```js
const { types, actions } = createActions([
  'ACTION_ONE',
  'ACTION_TWO',
])

// types   => { ACTION_ONE: 'ACTION_ONE', ... }
// actions => { actionOne: function() {}, ... }
```

Maybe something to consider along with the other examples in this thread as far as the final syntax.

yangmillstheory:
I can't see how this addition would be useful, especially with ES6 modules. 

You could get the same behavior with

```js
// actions/index.js
import { createAction } from 'redux-actions'

export const myFirstActionType = createAction('MY_FIRST_ACTION_TYPE')
export const mySecondActionType = createAction('MY_SECOND_ACTION_TYPE')

export default { myFirstActionType, mySecondActionType }
```

which is less code and indirection IMO.

jeffbski:
It's really about tightening up the code. Yes you can always create actions separately as you mentioned, we're not suggesting taking that away.

When you create lots of actions in a file, it becomes a big difference. Just increasing this to 6 items, results in much more code.

```js
export const actions = createActions('ONE', 'TWO', 'THREE', 'FOUR', 'FIVE', 'SIX');
```

compared to manual style

```js
const one = createAction('ONE');
const two = createAction('TWO');
const three = createAction('THREE');
const four = createAction('FOUR');
const five = createAction('FIVE');

export const actions = {
  one,
  two,
  three,
  four,
  five,
  six
};
```

It doesn't seem that it would be much of a cognitive stretch to have createAction and createActions. One simply implies that it does multiple in one call. 

It eliminates the need to remember to both create an action and add that action to the object. It can be easy to forget the second part.

yangmillstheory:
@jeffbski I generally agree that a pluralized version doesn't seem outlandish. I'm just not sure this belongs in the library, if it can easily be pushed out to userland.

For example, there's a way to structure your example (and mine) above such that you don't need to make changes in two places.

First pass, which assumes you don't need to do anything with the action creators themselves in the exporting module (like your first snippet)

```js
export const actions = {
  one: createAction('ONE'),
  two: createAction('TWO'),
  three: createAction('THREE'),
  four: createAction('FOUR'),
  five: createAction('FIVE'),
  six: createAction('SIX'),
};
```

Second approach, which assumes the exporting module needs to do something with the defined action creators

```js
const actions = {
  one: createAction('ONE'),
  two: createAction('TWO'),
  three: createAction('THREE'),
  four: createAction('FOUR'),
  five: createAction('FIVE'),
  six: createAction('SIX'),
};

// ...
// do whatever you want with actions object here
// ...

// this will never have to be updated when adding a new action
export const actions;
```

Are there other use cases I'm missing?

jeffbski:
The other interesting use case mentioned previously was to support an object which would allow for other optional tranformation (defaulting to identity if not a function).

```js
const actions = createActions({
 ONE: (foo, bar) => { a: foo, b: bar }, // payload creator
 TWO: null // just use identity
}
```

This just makes it easy to implement a bunch of action creators in a concise way (with or without payload transformation).

Nothing you can't do manually, but it cleans up the code nicely.

Eliminates the noise of calling createAction over and over again.

yangmillstheory:
@jeffbski that sounds reasonable. It seemed we were trying to solve a different use case with the PR description, but basically this is just syntactic sugar. Looking at the API, one would naturally wonder why there's no `createActions` method.

Back to the PR, does anyone feel that there's any compelling reason why we're actually transforming the action string? Also, this new function would need to be documented in the README as @alexgorbatchev mentioned.

src/createActions.js
```diff
@@ -0,0 +1,11 @@
+import ownKeys from './ownKeys';
+import createAction from './createAction';
+import camelize from './camelize';
+
+// creates a map of camelized actionTypes to the corresponding actions
+export default function createActions(actionTypes, transform = camelize) {
+  return ownKeys(actionTypes)
+    .reduce((acc, type) => (
+      Object.assign(acc, { [transform(type)]: createAction(type) })
```
yangmillstheory: Can you use the [object spread operator](http://redux.js.org/docs/recipes/UsingObjectSpreadOperator.html) here, as it seems to be the idiomatic way to do this in the Redux community?

In the [Babel REPL with our presets](https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-0&code=const%20o%20%3D%20%5B'a'%2C%20'b'%2C%20'c'%5D.reduce((memo%2C%20value)%20%3D%3E%20%7B%0A%20%20return%20%7B%20...memo%2C%20%5Bvalue%5D%3A%20value%20%2B%20'%20value'%20%7D%0A%7D%2C%20%7B%7D)%0A%0Aconsole.log(o))

```js
const o = ['a', 'b', 'c'].reduce((memo, value) => {
  return { ...memo, [value]: value + ' value' }
}, {})

console.log(o) // {"a":"a value","b":"b value","c":"c value"}
```

src/__tests__/createActions-test.js
```diff
@@ -0,0 +1,29 @@
...
+      MY_LONG_ACTION: 'MY_LONG_ACTION'
+    };
+
+    it(`apply a default string transformation function to action types
+        when mapping to action creators`, () => {
+      const actionCreators = createActions(types);
+
+      expect(actionCreators).to.have.property('action');
+      expect(actionCreators).to.have.property('myLongAction');
+    });
```
yangmillstheory: If I pass in an action `fooBar`, the behavior will be to make the action creator key `foobar`.

```js
function camelize(s) {
  return s.toLowerCase().replace(/_\w/g, x => x[1].toUpperCase())
}

console.log(camelize('fooBar')) // foobar
```

Is this the intended behavior?

src/createActions.js
yangmillstheory: Per @jeffbski, it would be great if this could support custom payload creators.

yangmillstheory:
@jeffbski do you have any objections to using `undefined` instead of `null` to default to the identity? Implementation-wise, we could use ES6 default arguments to build the final action creator.

yangmillstheory:
How would this handle passing in `metaCreators`?

src/__tests__/createActions-test.js
alanrsoares: `camelize` is just an arbitrary function that assumes that you're using the conventions of `UPPERCASE_UNDERSCORE_CONSTANT_NAMES`. 
If you're using a different convention you're free to pass your own transformation function.

yangmillstheory: What should the output be if you camelize an already camelized string?

As it is, if you camelize FOO_BAR, and camelize the output of that, you get something different.

alanrsoares: Sorry, I didn't quite get your point with this question.

yangmillstheory: >Is this the intended behavior?

What 's confusing you here?

The reason I'm asking is that your camelize function behaves unexpectedly and surprisingly. I'd think this from the point of view of a library developer as well as a library consumer.

It looks like if I defined an action "myActionType", it would be translated to "myactiontype". This is strange, given the behavior for uppercase strings, and given that your function is called "camelize" (it should do nothing to an already camelized string). I shouldn't have to pass in my own "transformation" function, it should just work.

This seems to be one of several limitations with this pull request, IMO. Anything else I can help clarify for you?

jeffbski:
I wasn't sure if we wanted to extend functionality to include metaCreators. I kind of figured that maybe this is a less used feature and if they wanted to use that they would resort to the standard createAction.

If we wanted to support metaCreators in createActions too, then we could do so by having one pass an object instead of a function. If they pass a function then it is assumed to be a payloadCreator, the most common need. If they pass an object then they can define payloadCreator and/or metaCreator functions.

```js
const actions = createActions({
  'ONE': (foo, bar) => ({ a: foo, b: bar }), // fn indicates payload creator
  'TWO': undefined, // use default payload creator
  'THREE': {  // use object to specify payloadCreator and/or metaCreator
    payload: (cat, dog) => ({ c: cat, d: dog }),
    meta: (cat, dog) => ({ id: baz })
  }
});

// Or if they only needed all identity actionCreators one can use
const actions = createActions(['ONE', 'TWO', 'THREE']);
```
I wasn't sure it would be better to use payload, payloadCreator, payloadFn for the object keys (and same for meta). Need to balance amount of typing with clarity. 

I think this above provides nice solutions for common use cases.

1. when you only need identity action creators, just pass an array of types and you are done
2. If you need payload creator on one or more then use an object and pass the payloadCreator as a function value or undefined to use identity
3. If you need meta creators then use an object and use an object for the value with payload and/or meta creators

The more common use cases are very succinct while we still intuitively support the full set of features too.

yangmillstheory:
I feel like this should also be able to declare many actions via the `next`/`throw` style.

jeffbski:
@yangmillstheory I'm not sure if I understand what you are suggesting with relation to next / throw style regarding actions. I see that in the action handlers.

Are you suggesting that handleActions would be able to automatically create actions as well as define the action handler reducer?

Or that createActions could define the actions and the handlers?

Maybe an example code snippet would help to explain what you were suggesting.

Thanks.

yangmillstheory:
@jeffbski I was talking nonsense; you're right, `next`/`throw` are in the action handlers.

The last thing that I would suggest is notation like https://github.com/yangmillstheory/redux-standard-actions/pull/1/files#diff-51e95fb5e086002b1e0cb0c0e77981bdR85, where we allow trailing action types which are translated to action creators with default payload creators.

I don't think I'm planning on submitting a PR for the work I've done in my fork, by the way. There's a lot of issues in this repo (naming, new features, better dev build system), that I think I can get done fairly quickly and this project seems to be lacking active maintenance. O course the code in my fork is free for use by anyone.

jeffbski:
@yangmillstheory that seems nice so you can easily create both actions with payload (and/or meta creators) along with simple identity actions. I like it.

So just for clarity here's what we have so far:

```js
const actions = createActions({
  // these actions need special payload and/or meta creators
  'ONE': (foo, bar) => ({ a: foo, b: bar }), // fn indicates payload creator
  'THREE': {  // use object to specify payloadCreator and/or metaCreator
    payload: (cat, dog) => ({ c: cat, d: dog }),
    meta: (cat, dog) => ({ id: baz })
  }
}, // any number of strings here create default identity action creators for these types
  'TWO',
  'FOUR'
);

// if you don't need any special payload/meta creators, then just the skip the first object and 
// provide action type strings
const actions = createActions('ONE', 'TWO', 'THREE');
```

So this makes it easy to create any number of identity actions as well as actions that need special payload or meta creators. The first object (specifying actions with custom payload and/or meta creators) is optional.

yangmillstheory:
Yep, that's right.

jeffbski:
I created a new pull request #104 to implement the ideas we have discussed here. See what you think.

timche:
#104 provides an improved version of this feature, let's move the discussion there.

PR #85, opened 2016-04-28
Adds TypeScript typings

alexgorbatchev:
This PR adds TypeScript typings to the packages without changes to the source files (I'd also be open to make a PR to convert the code base to TypeScript).

There's now `make test-typings` or `npm run test-typings` that will install `npm install typescript` if it's not installed locally already and run a test. I decided not to make this a part of a regular test in order to avoid making TypeScript a package dependency, but I'm happy to update and make it run all together.

What do you think?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,10 @@ clean:
 test: lint
 	NODE_ENV=test $(BIN)/mocha $(MOCHA_ARGS) $(MOCHA_TARGET)
 
+test-typings:
+	if [ ! -f $(BIN)/tsc ]; then npm install typescript; fi
+	$(BIN)/tsc --noEmit --noImplicitAny src/__tests__/typescript-typings-test.ts
+
 test-watch: lint
 	NODE_ENV=test $(BIN)/mocha $(MOCHA_ARGS) -w $(MOCHA_TARGET)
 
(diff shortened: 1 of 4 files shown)


alexgorbatchev:
@timche what do you think?

timche:
Hey, sorry. I had absolutely no time the last weeks.

Typings are definitely a good thing, but I'm not sure if TypeScript is the right choice. Many people are also going for Flow, so I think we should have a discussion which of them will be favoured.

timche:
As we are going with Flowtype in [v1.0.0](https://github.com/acdlite/redux-actions/issues/122), I will close this PR.

DataWingSoftware:
For those of us that do use TypeScript can you help us understand the detriment to your code base by adding this? It doesn't seem like it would harm your code, and it would make it more open and inclusive for use by the broader community.

alexgorbatchev:
@DataWingSoftware this isn't a huge issue. If there's a desire to have this all be packaged in one, we can always publish a typed fork.

stephenlaughton:
What is the real project overhead of including Typescript typings? Someone who is using Typescript can maintain them and they are out of sight/mind for anyone who isn't using them. I agree with @DataWingSoftware on this one. @alexgorbatchev it seems like more overhead to maintain a Typescript fork?

PR #97, opened 2016-06-28
Simplified implementation of handleAction

Lucretiel:
- Moved the function-or-object check on the `reducers` parameter to outside the closure function
- Removed the use of magic strings
- Removed the mutation of the passed-in `reducers` argument

### DIFF
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -2,6 +2,10 @@ function isFunction(val) {
   return typeof val === 'function';
 }
 
+function coerceReducer(reducer) {
+  return isFunction(reducer) ? reducer : (state, action) => state
+}
+
 export default function handleAction(type, reducers, defaultState) {
   const typeValue = isFunction(type)
     ? type.toString()
@@ -9,13 +13,8 @@ export default function handleAction(type, reducers, defaultState) {
 
   const [nextReducer, throwReducer] = isFunction(reducers)
     ? [reducers, reducers]
-    : [reducers.next, reducers.throw];
-
-  if (nextReducer === undefined || throwReducer === undefined) {
-    throw new TypeError(
-      'reducers argument must be either a function or an object of shape {next, throw}'
-    );
-  }
+    : [reducers.next, reducers.throw].map(coerceReducer);
+  // TODO: warn if both reducers are not functions; that would make this a no-op reducer.
 
   return (state = defaultState, action) => {
     // If action type does not match, return previous state
@@ -23,8 +22,6 @@ export default function handleAction(type, reducers, defaultState) {
 
     const reducer = action.error === true ? throwReducer : nextReducer;
 
-    return isFunction(reducer)
-      ? reducer(state, action)
-      : state;
+    return reducer(state, action)
   };
 }
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -2,8 +2,8 @@ function isFunction(val) {
   return typeof val === 'function';
 }
 
-function coerceReducer(reducer) {
-  return isFunction(reducer) ? reducer : state => state;
+function identity(state) {
+  return state;
 }
 
 export default function handleAction(type, reducers, defaultState) {
@@ -11,10 +11,22 @@ export default function handleAction(type, reducers, defaultState) {
     ? type.toString()
     : type;
 
-  const [nextReducer, throwReducer] = isFunction(reducers)
+  const [nextReducer = identity, throwReducer = identity] = isFunction(reducers)
     ? [reducers, reducers]
-    : [reducers.next, reducers.throw].map(coerceReducer);
-  // TODO: warn if both reducers are not functions; that would make this a no-op reducer.
+    : [reducers.next, reducers.throw];
+  // TODO: warn if both reducers are undefined or identity; that would make this a no-op reducer.
+
+  // TODO: relace this kludge with flow or some other proper type checker. 
+  if (!isFunction(nextReducer)) {
+    throw new TypeError(
+      "If given, reducers or reducers.next must be a function (got " + nextReducer + ")"
+    );
+  }
+  if (!isFunction(throwReducer)) {
+    throw new TypeError(
+      "If given, reducers or reducers.throw must be a function (got " + throwReducer + ")"
+    );
+  }
 
   return (state = defaultState, action) => {
     // If action type does not match, return previous state
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -12,7 +12,9 @@ export default function handleAction(type, reducers, defaultState) {
     : [reducers.next, reducers.throw];
 
   if (nextReducer === undefined || throwReducer === undefined) {
-    throw TypeError("reducers argument must be either a function or an object of shape {next, throw}");
+    throw new TypeError(
+      'reducers argument must be either a function or an object of shape {next, throw}'
+    );
   }
 
   return (state = defaultState, action) => {
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -7,19 +7,14 @@ export default function handleAction(type, reducers, defaultState) {
     ? type.toString()
     : type;
 
+  const nextReducer = isFunction(reducers) ? reducers : reducers.next;
+  const throwReducer = isFunction(reducers) ? reducers : reducers.throw;
+
   return (state = defaultState, action) => {
     // If action type does not match, return previous state
     if (action.type !== typeValue) return state;
 
-    const handlerKey = action.error === true ? 'throw' : 'next';
-
-    // If function is passed instead of map, use as reducer
-    if (isFunction(reducers)) {
-      reducers.next = reducers.throw = reducers;
-    }
-
-    // Otherwise, assume an action map was passed
-    const reducer = reducers[handlerKey];
+    const reducer = action.error === true ? throwReducer : nextReducer;
 
     return isFunction(reducer)
       ? reducer(state, action)
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -7,8 +7,9 @@ export default function handleAction(type, reducers, defaultState) {
     ? type.toString()
     : type;
 
-  const nextReducer = isFunction(reducers) ? reducers : reducers.next;
-  const throwReducer = isFunction(reducers) ? reducers : reducers.throw;
+  const [nextReducer, throwReducer] = isFunction(reducers)
+    ? [reducers, reducers]
+    : [reducers.next, reducers.throw];
 
   return (state = defaultState, action) => {
     // If action type does not match, return previous state
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -3,7 +3,7 @@ function isFunction(val) {
 }
 
 function coerceReducer(reducer) {
-  return isFunction(reducer) ? reducer : (state, action) => state
+  return isFunction(reducer) ? reducer : state => state;
 }
 
 export default function handleAction(type, reducers, defaultState) {
@@ -22,6 +22,6 @@ export default function handleAction(type, reducers, defaultState) {
 
     const reducer = action.error === true ? throwReducer : nextReducer;
 
-    return reducer(state, action)
+    return reducer(state, action);
   };
 }
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -11,6 +11,10 @@ export default function handleAction(type, reducers, defaultState) {
     ? [reducers, reducers]
     : [reducers.next, reducers.throw];
 
+  if (nextReducer === undefined || throwReducer === undefined) {
+    throw TypeError("reducers argument must be either a function or an object of shape {next, throw}");
+  }
+
   return (state = defaultState, action) => {
     // If action type does not match, return previous state
     if (action.type !== typeValue) return state;
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -16,15 +16,15 @@ export default function handleAction(type, reducers, defaultState) {
     : [reducers.next, reducers.throw];
   // TODO: warn if both reducers are undefined or identity; that would make this a no-op reducer.
 
-  // TODO: relace this kludge with flow or some other proper type checker. 
+  // TODO: relace this kludge with flow or some other proper type checker.
   if (!isFunction(nextReducer)) {
     throw new TypeError(
-      "If given, reducers or reducers.next must be a function (got " + nextReducer + ")"
+      `If given, reducers or reducers.next must be a function (got ${nextReducer})`
     );
   }
   if (!isFunction(throwReducer)) {
     throw new TypeError(
-      "If given, reducers or reducers.throw must be a function (got " + throwReducer + ")"
+      `If given, reducers or reducers.throw must be a function (got ${throwReducer})`
     );
   }
 
--- a/README.md
+++ b/README.md
@@ -93,6 +93,15 @@ handleAction('FETCH_DATA', {
 });
 ```
 
+If you leave either reducer unspecified, the wrapped reducer will treat that reducer as a no-op, returning the state unchanged.
+
+```js
+handleAction('FETCH_DATA', {
+  next(state, action) {...}
+  // State will be unchanged in the event of an error
+});
+```
+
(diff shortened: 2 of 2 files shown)


src/handleAction.js
```diff
@@ -7,19 +7,14 @@ export default function handleAction(type, reducers, defaultState) {
     ? type.toString()
     : type;
 
+  const nextReducer = isFunction(reducers) ? reducers : reducers.next;
+  const throwReducer = isFunction(reducers) ? reducers : reducers.throw;
```
yangmillstheory: What do you think about using multiple assignment to DRY things up here?

In the [Babel REPL with our `.babelrc`](https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-0&code=let%20foo%0Alet%20bar%0Aconst%20o%20%3D%20%7B%20foo%3A%20'baz'%2C%20bar%3A%20'qux'%20%7D%0A%0Aif%20(false)%20%7B%0A%20%20%5B%20foo%2C%20bar%20%5D%20%3D%20%5B%20o.foo%2C%20o.bar%20%5D%20%0A%7D%20else%20%7B%0A%20%20%5B%20foo%2C%20bar%20%5D%20%3D%20%5B%20o%2C%20o%20%5D%20%0A%7D%0A%0A%0Aconsole.log(foo%2C%20bar)):

```js
let foo
let bar
const o = { foo: 'baz', bar: 'qux' }

if (flag) {
  [ foo, bar ] = [ o.foo, o.bar ] 
} else {
  [ foo, bar ] = [ o, o ] 
}

// flag === true => baz qux
// flag === false =>  {"foo":"baz","bar":"qux"} {"foo":"baz","bar":"qux"}
console.log(foo, bar) 
```

so we could be more DRY here with

```js
let nextReducer = throwReducer = reducers;

if (!isFunction(reducers)) {
  [ nextReducer, throwReducer ] = [ reducers.next, reducers.throw ]
}
```

src/handleAction.js
```diff
@@ -7,19 +7,14 @@ export default function handleAction(type, reducers, defaultState) {
...
 
-    const handlerKey = action.error === true ? 'throw' : 'next';
-
-    // If function is passed instead of map, use as reducer
-    if (isFunction(reducers)) {
-      reducers.next = reducers.throw = reducers;
-    }
-
-    // Otherwise, assume an action map was passed
-    const reducer = reducers[handlerKey];
```
yangmillstheory: Is the idea of this to move the logic up to the proper scope?

src/handleAction.js
Lucretiel: Oh that's perfect. I was actually wishing I could do that but I couldn't work out a clean way to do it... I didn't know you could do `[a, b] = ...` in addition to `{a, b} = ...` Will update, probably with:

```
const [nextReducer, throwReducer] = isFunction(reducers) ?
    [reducers, reducers] :
    [reducers.next, reducers.throw]
```

src/handleAction.js
Lucretiel: Couple ideas. The main one was, yeah, to move logic as high up in the scope as possible. I also wanted to get rid of the magic strings `'throw'` and `'next'`, the property accessor `reducers[handlerKey]` in favor of the more explicit `nextReducer`/`throwReducer`, and generally short circuit the whole process as much as possible. The fact that the new version has fewer lines of code and is (imo) simpler is an unintended but welcome side effect.

src/handleAction.js
```diff
@@ -7,19 +7,15 @@ export default function handleAction(type, reducers, defaultState) {
...
+    const reducer = action.error === true ? throwReducer : nextReducer;
 
     return isFunction(reducer)
       ? reducer(state, action)
```
yangmillstheory: I was thinking this might not be necessary with your changes, but I guess either `reducers.throw` or `reducers.next` might not be functions.

I was wondering if we should fail earlier if `!isFunction(reducers)` and `reducers` is not an object with function properties `next` and `throw` (not necessarily suggesting it be implemented in this PR). It would eliminate the downstream checks.

yangmillstheory:
👍  nice improvement of hygiene. ping @timche

src/handleAction.js
Lucretiel: I can add that; this PR actually makes that more natural anyway. Check to make sure that `nextReducer` and `throwReducer` `!== undefined`, and throw otherwise?

Lucretiel: We should assume that doing something like this is an error, write?

```
const reducer = handleAction(myAction, {
    next: ...
    // No throw reducer
})

File: src/handleAction.js [hunk 4]
```diff
@@ -7,19 +7,15 @@ export default function handleAction(type, reducers, defaultState) {
...
-    const handlerKey = action.error === true ? 'throw' : 'next';
-
-    // If function is passed instead of map, use as reducer
-    if (isFunction(reducers)) {
-      reducers.next = reducers.throw = reducers;
-    }
-
-    // Otherwise, assume an action map was passed
-    const reducer = reducers[handlerKey];
+    const reducer = action.error === true ? throwReducer : nextReducer;
```
Lucretiel: Just a quick note: I double checked the FSA spec, and confirmed that this `=== true` behavior is in fact the most correct (as opposed to doing something like `action.error ? throwReducer : nextReducer`):

> If error has any other value besides true, including undefined and null, the action MUST NOT be interpreted as an error.

So 👍 to whoever got that right.

src/handleAction.js
yangmillstheory: I think this is a case of a breaking change, but it's also the library doing what the documentation says

>If a single reducer is passed, it is used to handle both normal actions and failed actions. (A failed action is analogous to a rejected promise.) You can use this form if you know a certain type of action will never fail, like the increment example above.
>
> Otherwise, you can specify separate reducers for next() and throw(). This API is inspired by the ES6 generator interface.

So it seems it doesn't make sense to not pass `throw` and just pass `next` in the object form, since the former case would handle that.

For your implementation, why not do an `isFunction` check instead of checking `undefined`? 

Also, you could ensure that you can dot into `reducers` first before checking the types of `reducers.throw` and `reducers.next`. [This would have thrown errors previously](https://github.com/acdlite/redux-actions/blob/master/src/handleAction.js#L21), so it's not a breaking change, we'd just be being more explicit about why it failed here.

src/handleAction.js
yangmillstheory: Cool, I was going to ask why we weren't checking truthiness, but that settles that.

src/handleAction.js
yangmillstheory: Another positive effect of this change would be that `handleAction` would fail earlier at the definition level, and not the invocation level. Failing earlier is 👍 .

src/handleAction.js
yangmillstheory: I like that these lines of code isn't run every time the action is being handled.

src/handleAction.js
yangmillstheory: Note that these checks would make a use case of factoring out https://github.com/acdlite/redux-actions/blob/master/src/handleAction.js#L1 into its own module, like [`ownKeys.js`](https://github.com/acdlite/redux-actions/blob/master/src/ownKeys.js).

Lucretiel: Added a patch; you can tell me what you think. We're definitely into a place where my grasp of precise javascript semantics is pretty fuzzy.

As for the breaking change thing– I had the same question about a python library. It turns out that the documentation is really the only source of truth; changing behavior that is undocumented, or errors thrown because of use the library incorrectly or in an undocumented way, isn't considered making a breaking change. I'll leave it up to you guys whether you think it counts in this case.

See also the question I asked here: https://github.com/mojombo/semver/issues/248. Basically, I was worried that adding new arguments to a python function would constitute a breaking change, because people who were calling the function with those arguments in a prior version and seeing a `TypeError` wouldn't see that anymore.

Lucretiel: > Note that these checks would make a use case of factoring out...

That reminds me– are we trying to keep this project minimally dependent on others? Because lodash has good implementations of both (isFunction)[https://lodash.com/docs#isFunction] and (ownKeys)[https://lodash.com/docs#keys] we could use.

yangmillstheory: Yeah, it would seem that it's explicitly trying to not take dependencies. There's also https://github.com/acdlite/redux-actions/blob/master/src/createAction.js#L1 which is included in `_`. 

IMO we could quite easily factor out `isFunction` though, especially since it's already in the codebase. I think this would be more preferable than taking a dependency.

I've learned that forcing library users to take dependencies can be off-putting. So I've explicitly removed them where I could from my own libraries (in my case people were complaining about `_` and bundle size).

src/handleAction.js
```diff
@@ -12,7 +12,9 @@ export default function handleAction(type, reducers, defaultState) {
     : [reducers.next, reducers.throw];
 
   if (nextReducer === undefined || throwReducer === undefined) {
-    throw TypeError("reducers argument must be either a function or an object of shape {next, throw}");
+    throw new TypeError(
+      'reducers argument must be either a function or an object of shape {next, throw}'
```
yangmillstheory: >reducers must be a reducer function or an object with shape {next, throw} where 'next' and 'throw' are reducer functions

I know you were broken with ESLint for long lines, so you could use string concatenation:

```js
throw new TypeError(
  'reducers must be a reducer function or an object with ' + 
  'shape {next, throw} where 'next' and 'throw' are reducer functions'
)
```

Lucretiel:
Hmmm. So the tests that are failing are, in fact, using the partial object style:

```
it('uses `next()` if action does not represent an error', () => {
  const reducer = handleAction(type, {
    next: (state, action) => ({
      counter: state.counter + action.payload
    })
  });
  expect(reducer(prevState, { type, payload: 7 }))
    .to.deep.equal({
      counter: 10
    });
});
```

The documentation makes no mention of this style being valid, and in practice I can't think of any reasonable behavior for it. Silently skipping actions that are (or aren't) errors seems like an accident waiting to happen, and we already have a syntax for a unified reducer. It's my belief, therefore, that this test and its ilk are buggy, and there should at least be a placeholder reducer added to make it pass. However, I'll await advice from the rest of you before making this change.

src/handleAction.js
Lucretiel: That's fine; it runs against my personal grain, but it's a perfectly valid standpoint. Doing that refactoring seems outside the scope of this PR, though, so we'll do it after this is resolved.

yangmillstheory:
Looked at the Travis failures and I agree that placeholder reducers (https://github.com/acdlite/redux-actions/blob/master/src/createAction.js#L1) should be added to those tests. Thinking of tests as a form of API documentation, this makes even more sense. Does adding them make the tests pass?

Lucretiel:
Oh, dear. It actually appears that, based on the tests (thought not the written documentation), the defined behavior is in fact to do nothing if no function is attached to either the `next` or `throw` keys:

https://github.com/acdlite/redux-actions/blob/master/src/__tests__/handleAction-test.js#L87-L92

There is no similar test in place for passing a null as the reducer argument itself. This, in fact, was changed from the prior behavior, which DID raise an exception if a function wasn't passed in: https://github.com/Lucretiel/redux-actions/commit/e6591594ba275c001c19390bf4d575adba6455f1#diff-13523acaa823765ceef717519f2dcc3dL75

It's not clear to me if this change was made as an explicit API change, or if it was just changed with some subtle javascript thing. Pinging @acdlite for his input, as he wrote the change.

It's one thing for it to be a subtle side effect; it's another thing entirely if there's an explicitly described test describing the behavior. At the time, the addition of this (and possibly other) changes were seen as only needing a [minor version bump](https://github.com/Lucretiel/redux-actions/commit/e6591594ba275c001c19390bf4d575adba6455f1). Semver itself leads it up to our interpretation what counts as the "documented public API." It's still my preference that this behavior be changed to raise as error, especially because that error would be raised from `handleAction` and not from the returned action creator. It's also my preference that, because the written documentation in the README does say either way, we consider this only a patch change, or perhaps a new minor feature because of the added type checking. But I'll defer to the wisdom of the rest of you on that.

src/handleAction.js
```diff
@@ -7,19 +7,21 @@ export default function handleAction(type, reducers, defaultState) {
...
+
   return (state = defaultState, action) => {
     // If action type does not match, return previous state
     if (action.type !== typeValue) return state;
 
-    const handlerKey = action.error === true ? 'throw' : 'next';
-
-    // If function is passed instead of map, use as reducer
-    if (isFunction(reducers)) {
-      reducers.next = reducers.throw = reducers;
```
yangmillstheory: FYI, on my fork with stricter ESLint rules, this fails with 

```
/Users/VictorAlvarez/code/private/redux-actions/src/handleAction.js
  18:7   error  Assignment to property of function parameter 'reducers'  no-param-reassign
  18:23  error  Assignment to property of function parameter 'reducers'  no-param-reassign
```

src/handleAction.js
yangmillstheory: On my [fork](https://github.com/yangmillstheory/redux-standard-actions/pull/2/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L36), I've started using lodash micro-packages. 

Since I'm making faster progress there, I'm considering renaming it and publishing to NPM.

src/handleAction.js
Lucretiel: Yeah. That was the original problem I saw that compelled me to make this fix– mutating/assigning to the input parameter seemed like a VERY bad idea.

Lucretiel:
So, after sleeping on it, I decided that in this case it's best to defer to the behavior explicitly described by the tests. I also decided that only specifying one of `next` or `throw` could make sense in certain situations, like when you want to ignore errors. With this in mind, I reverted the type check, and added some extra supporting code to further do as much preprocessing of the reducers as possible, minimizing the work done by the actual returned reducer. I also added some extra docs to the README about the optional reducers.

yangmillstheory:
Nice. Does the case of passing in a reducer function as opposed to an object handle the case of wanting to ignore errors?

Lucretiel:
Not by itself, no. As per the docs:

> If a single reducer is passed, it is used to handle both normal actions and failed actions.

If you pass a single reducer function, instead of an object, the function is invoked for all objects of that type.

Lucretiel:
Keep in mind that that's the typical case anyway. Most users won't ever have a need to create or handle error actions. I myself am already not a fan of async or thunk action creators, because they don't give you a useful handle to the in-progress action that you can cancel or do something else with.

yangmillstheory:
Yes, I've not found a good use case for this myself, yet.

src/handleAction.js
```diff
@@ -2,27 +2,26 @@ function isFunction(val) {
   return typeof val === 'function';
 }
 
+function coerceReducer(reducer) {
+  return isFunction(reducer) ? reducer : state => state;
```
yangmillstheory: This isn't a blocking concern, but I still wish we were more intentional about the argument type passed in. 

`null` and `undefined` seem reasonable to coerce to a default, but `1`, `{}`, `[]`, `true`, `false` seems like a misuse of the API that we should throw `TypeError` on.

IMO, the most ideal situation would be to not conflate `null` and `undefined`, and just pick one to coerce, preferably `undefined`, so we could use default argument syntax like https://github.com/yangmillstheory/redux-standard-actions/blob/master/src/actionReducer.js#L27.

yangmillstheory:
@timche could you take a look?

@Lucretiel do you mind if I push some minor formatting changes to your branch?

Thanks for your patience. I'd love to get this small PR (that's generated some good useful discussion) in soon!

src/handleAction.js
```diff
@@ -2,19 +2,31 @@ function isFunction(val) {
   return typeof val === 'function';
```
yangmillstheory: Sorry, https://github.com/acdlite/redux-actions/pull/97#discussion_r70165690 wasn't meant to suggest this change - it's more of a longer-term vision for this library. You can revert this commit if you like.

(Also, you can use ES6 string interpolation below.)

yangmillstheory:
This has gotten a bit stale and has got tons of discussion, let's move this over to #109.

Issue #101, opened 2016-07-02
Parameter naming in createAction doesn't make sense


yangmillstheory:
Generally related to #47.

https://github.com/acdlite/redux-actions/blob/master/src/createAction.js#L5 calls the payload creator `actionCreator`, which is a very different thing.

Actually the [signature of `createAction` in the README](https://github.com/acdlite/redux-actions/blob/master/README.md#createactiontype-payloadcreator--identity-metacreator) spells out pretty clearly what this parameter should be called.

timche:
Would you mind opening a PR to change this?

yangmillstheory:
Closed in https://github.com/acdlite/redux-actions/commit/e7bdb2f6109ec196ce097bccd7026bf1497a1643.

PR #105, opened 2016-07-09
Fix payload creator naming, close #103, #101.

### LINKED ISSUE #103: Tests method names wrongly call payload creators "action creators"
This adds to already confusing terminology (#47, #101).

The following lines should be changed to stop calling payload creators "action creators":

- https://github.com/acdlite/redux-actions/blob/master/src/__tests__/createAction-test.js#L120
- https://github.com/acdlite/redux-actions/blob/master/src/__tests__/createAction-test.js#L25

yangmillstheory:
See #101, #103.

### DIFF
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -22,7 +22,7 @@ describe('createAction()', () => {
       });
     });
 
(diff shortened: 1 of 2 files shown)


timche:
Thanks! :) 👍

PR #107, opened 2016-07-09
Refactor Build System

timche:
@yangmillstheory can you review it please? :)


package.json
```diff
@@ -4,11 +4,16 @@
   "description": "Flux Standard Action utlities for Redux",
   "main": "lib/index.js",
   "scripts": {
-    "test": "make test",
-    "prepublish": "make clean build"
+    "build": "npm run clean && babel src --out-dir lib --ignore *-test.js",
```
yangmillstheory: Will this build __tests__/init.js?

yangmillstheory:
Looks pretty elegant. I'll take a look today. :+1:

package.json
timche: Nope, or what do you mean by that?

yangmillstheory: Ah, I thought `__tests__/init.js` still existed, but I see it's been moved to `test/init.js`.

What do you think about moving the contents of `__tests__` to `test`, to consolidate (seems like test-related files are in two places right now)? This would simplify the build script to

```js
"build": "npm run clean && babel src --out-dir lib",
```

and the test scripts would look like

```js
"test": "mocha --compilers js:babel-register --require test/init.js test",
"test:watch": "npm run test -- --watch test"
```

The test files would import from `src/`:

```js
import { handleAction, createAction } from '../src';
```

package.json
```diff
@@ -29,16 +34,19 @@
...
     "babel-register": "^6.7.2",
     "chai": "^3.0.0",
     "eslint": "^2.8.0",
     "eslint-config-airbnb-base": "^1.0.3",
     "eslint-plugin-import": "^1.5.0",
-    "lodash.isplainobject": "^3.2.0",
+    "eslint-watch": "^2.1.13",
     "flux-standard-action": "^0.6.0",
-    "mocha": "^2.2.5"
+    "lodash.isplainobject": "^3.2.0",
```
yangmillstheory: I don't think this is actually being used - maybe we should get rid of it now as getting rid of unused `devDependencies` might fall under refactoring the local build/dev workflow?

yangmillstheory:
I've tested this out locally:

- [x] failed linting exits with an error
- [x] failed tests exist with an error
- [x] continuous linting works
- [x] continuous testing works
- [x] build and clean work
- [x] prepublish should work based on all of the above

So I just have two questions, otherwise looks great. This makes me want to start using NPM scripts more.

package.json
timche: You are right. I've removed it.

package.json
timche: The `test` folder is only to setup the test environment. Personally I like the approach of having a `__tests__` subfolder which is mirroring the structure of it's parent folder.

Another solution could be to remove the `init.js` and import `chai` in the test files.

yangmillstheory: I like the idea of `import 'chai'` - no need for extra build complexity just to set up that one global. It would also help clarify what assertion library we're using.

timche: Resolved with 64b6a51.

.eslintignore
```diff
@@ -0,0 +1 @@
+lib
```
yangmillstheory: Is it possible to lint `src/` directly [now that `test/` is gone](https://github.com/acdlite/redux-actions/commit/64b6a51bdc024d4d0af1dd67017a274881d785ef)?

timche: Nice, catch. Changed it with e8cc5de.

yangmillstheory:
Hm, the Travis build has been running for quite some time now...

timche:
Travis is having some issues: https://www.traviscistatus.com/

yangmillstheory:
:shipit:

timche:
The checks have passed, may I merge? 😬

yangmillstheory:
Yes, please! The :shipit: emoji is usually the GitHub way to indicate an up vote for merge/deploy.

PR #109, opened 2016-07-14
Improved implementation of handleAction - Take 2

yangmillstheory:
This is based off of the ideas in #97, which generated a lot of good discussion. Roughly:

- don't mutate the passed in `reducers`
- use small lodash libraries
- properly scope reducer setup (i.e. do it once, and not every time an action is handled)

Full credit goes to @Lucretiel; just submitting it here for a cleaner pull request.

@Lucretiel, please review as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,10 +48,7 @@
     "rimraf": "^2.5.3"
   },
   "dependencies": {
-    "lodash.identity": "^3.0.0",
-    "lodash.isfunction": "^3.0.8",
-    "lodash.isnil": "^4.0.0",
-    "lodash.issymbol": "^4.0.1",
+    "lodash": "^4.13.1",
     "reduce-reducers": "^0.1.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -48,6 +48,10 @@
     "rimraf": "^2.5.3"
   },
   "dependencies": {
+    "lodash.identity": "^3.0.0",
+    "lodash.isfunction": "^3.0.8",
+    "lodash.isnil": "^4.0.0",
+    "lodash.issymbol": "^4.0.1",
     "reduce-reducers": "^0.1.0"
   }
 }
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -1,7 +1,7 @@
-import isFunction from 'lodash.isfunction';
-import identity from 'lodash.identity';
-import isNil from 'lodash.isnil';
-import isSymbol from 'lodash.issymbol';
+import isFunction from 'lodash/isfunction';
+import identity from 'lodash/identity';
+import isNil from 'lodash/isnil';
+import isSymbol from 'lodash/issymbol';
 
 export default function handleAction(type, reducers, defaultState) {
   const typeValue = isSymbol(type)
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -1,7 +1,7 @@
-import isFunction from 'lodash/isfunction';
+import isFunction from 'lodash/isFunction';
 import identity from 'lodash/identity';
-import isNil from 'lodash/isnil';
-import isSymbol from 'lodash/issymbol';
+import isNil from 'lodash/isNil';
+import isSymbol from 'lodash/isSymbol';
 
 export default function handleAction(type, reducers, defaultState) {
   const typeValue = isSymbol(type)
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -1,28 +1,22 @@
-function isFunction(val) {
-  return typeof val === 'function';
-}
+import isFunction from 'lodash.isfunction';
+import identity from 'lodash.identity';
+import isNil from 'lodash.isnil';
+import isSymbol from 'lodash.issymbol';
 
 export default function handleAction(type, reducers, defaultState) {
-  const typeValue = isFunction(type)
-    ? type.toString()
-    : type;
+  const typeValue = isSymbol(type)
+    ? type
+    : type.toString();
+
+  const [nextReducer, throwReducer] = isFunction(reducers)
+    ? [reducers, reducers]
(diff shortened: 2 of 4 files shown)


src/handleAction.js
```diff
@@ -1,28 +1,22 @@
...
-  const typeValue = isFunction(type)
-    ? type.toString()
-    : type;
-
-  return (state = defaultState, action) => {
-    // If action type does not match, return previous state
-    if (action.type !== typeValue) return state;
+  const typeValue = isSymbol(type)
+    ? type
+    : type.toString();
```
yangmillstheory: This works with string action types and action creators. I think calling out `isSymbol` explicitly clarifies the special case (where we cannot call `.toString()`).

package.json
```diff
@@ -48,6 +48,10 @@
     "rimraf": "^2.5.3"
   },
   "dependencies": {
+    "lodash.identity": "^3.0.0",
+    "lodash.isfunction": "^3.0.8",
+    "lodash.isnil": "^4.0.0",
+    "lodash.issymbol": "^4.0.1",
```
yangmillstheory: This would be a new change, but I think it's a good one. Thoughts? 

Willing to roll our own to minimize dependencies, but why re-invent the wheel?

yangmillstheory:
ping @timche for the new dependencies and that general approach
ping @Lucretiel

This is a small PR (from the long-outstanding #97) so, I'm planning to get merge sometime over the weekend. If there's no time to review, we can review retrospectively and I can address those comments after merge.

package.json
timche: Using `lodash` is fine, but I'd suggest to just go with `lodash` as dependency and use e.g. `import identity from 'lodash/identity'` instead. It has same effect but without polluting the `package.json` and we have access to all `lodash` libs if needed without installing new deps.

src/handleAction.js
timche: 👍

src/handleAction.js
```diff
@@ -1,28 +1,22 @@
...
+      return state;
     }
 
-    // Otherwise, assume an action map was passed
-    const reducer = reducers[handlerKey];
-
-    return isFunction(reducer)
-      ? reducer(state, action)
-      : state;
+    return (action.error === true ? throwReducer : nextReducer)(state, action);
```
timche: `=== true` could be omitted.

src/handleAction.js
```diff
@@ -1,28 +1,22 @@
...
-    // If action type does not match, return previous state
-    if (action.type !== typeValue) return state;
+  const typeValue = isSymbol(type)
+    ? type
+    : type.toString();
 
-    const handlerKey = action.error === true ? 'throw' : 'next';
+  const [nextReducer, throwReducer] = isFunction(reducers)
+    ? [reducers, reducers]
+    : [reducers.next, reducers.throw].map(reducer => (isNil(reducer) ? identity : reducer));
```
timche: This block could come after L17, so `if (action.type !== typeValue)` we can `return state` earlier.

package.json
yangmillstheory: b5546b5f

src/handleAction.js
yangmillstheory: What are your thoughts on https://github.com/acdlite/redux-actions/pull/97#discussion_r69210523?

timche: You are right, `error` could be any other value than `true`. `=== true` adds more safety to respect FSA.

src/handleAction.js
yangmillstheory: Hm...which part were you thinking could come after L18? 

Our initial thought in moving this up the scope was https://github.com/acdlite/redux-actions/pull/97#discussion_r69203656.

timche: Alright, I've gone through that again. I understand your idea know behind this. Leave it like that ;)

timche:
After clarified all the things, this PR is very neat 👌🏻. You got my checkmark!

yangmillstheory:
Thanks, and big thanks to @Lucretiel (this is really his PR). 

There's a build break with b5546b5

```
/Users/VictorAlvarez/code/private/redux-actions/src/handleAction.js (3/0)
  ✖  1:24  Unable to resolve path to module 'lodash/isfunction'  import/no-unresolved
  ✖  3:19  Unable to resolve path to module 'lodash/isnil'       import/no-unresolved
  ✖  4:22  Unable to resolve path to module 'lodash/issymbol'    import/no-unresolved
```

I'll see what I can do to fix it; let me know if you have any ideas on it.

src/handleAction.js
yangmillstheory: No problem! I thought the same thing; thanks to @Lucretiel for doing the research and pointing it out.

yangmillstheory:
9a317f3 fixed the build. The interesting lesson was that the environments the tests were run in had a case-insensitive filesystem, including mine:

```
$ touch abc1
$ touch abC1
$ ls -l ab*
-rw-r--r--  1 VictorAlvarez  staff  0 Jul 16 02:05 abc1
```

yangmillstheory:
@timche what do you think of releasing this in v0.10.2?

timche:
I would make this part of v0.11.0 with a bunch of other non-breaking PRs. Should try to follow [semver](http://semver.org/).

yangmillstheory:
Squashed merged in 5e78f252ed778ec2118ed7f01fdc8e973330bdf6.

Would you consider this addition a bug fix or new functionality?

>Given a version number MAJOR.MINOR.PATCH, increment the:

>MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.
Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

PR #111, opened 2016-07-25
reducer will crash if action is not defined

neekey:
reducer should work normally whether state or action is not given.


timche:
Do you have an example case where this explicitly fails?

yangmillstheory:
>reducer should work normally whether state or action is not given.

Wait, why?

yangmillstheory:
Closing due to inactivity. If possible, please submit an issue to discuss the problem and use cases before proposing changes to ensure pull requests move more efficiently.

PR #112, opened 2016-07-25
make "defaultState" a real default state

neekey:
it is common that we use Reducer to set a default state structure for every related action, and since through handleActions there's no way to create a common initial state for every action, I suggest instead of just relying on the ES6's default parameter feature, we use defaultState as a real default state.

And if defaultState is a non-object value, the ES6's default parameter feature should work well. Otherwise we use `Object.assign()` to do the trick.

A simple user case will be that I am implementing a reducer for a form logic, my state will include all the form fields, like below:

```json
{
  "values": {}
}
```

and I might want to define default fields:

```json
{
  "values": {
    "email": null,
    "name": null
  }
}
```

And to handle actions like 'EMAIL_CHANGE', I can just write:

```js
handleAction('EMAIL_CHANGE', (state, action) => {
    state.values.email = action.payload.email;
    return state;
}, {
  values: {
    email: null,
    name: null
  }
});
```

otherwise, I maybe need to do initialization for the state inside the reducer:

```js
handleAction('EMAIL_CHANGE', (state, action) => {
    state.values = state.values || { email: null, name: null };
    state.values.email = action.payload.email;
    return state;
}, {
  values: {
    email: null,
    name: null
  }
});
```

### DIFF
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -2,6 +2,7 @@ import isFunction from 'lodash/isFunction';
 import identity from 'lodash/identity';
 import isNil from 'lodash/isNil';
 import isSymbol from 'lodash/isSymbol';
+import isPlainObject from 'lodash/isPlainObject';
 
 export default function handleAction(type, reducers, defaultState) {
   const typeValue = isSymbol(type)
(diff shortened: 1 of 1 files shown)


yangmillstheory:
@neekey please open an issue and put this request there with all the information so we can determine if it makes sense.

yangmillstheory:
In your snippets above, by the way, you **should not** be mutating the state object.

http://redux.js.org/docs/Troubleshooting.html

> Never mutate reducer arguments
>
> It is tempting to modify the state or action passed to you by Redux. Don’t do this!
>
> Redux assumes that you never mutate the objects it gives to you in the reducer. Every single time, you must return the new state object. Even if you don’t use a library like Immutable, you need to completely avoid mutation.

It's also not clear what the issue is, but I'm guessing it can be handled with plain reducer composition in conjunction with `combineReducers`.

Issue #20, opened 2015-08-24
ES6 Symbols break handleActions


billyjanitsch:
`handleActions` doesn't correctly match actions with types that are `Symbol`s, e.g.

```js
const TEST_ACTION = Symbol('TEST_ACTION')
const OTHER_TEST_ACTION = 'OTHER_TEST_ACTION'
const reducer = handleActions({
  [TEST_ACTION]: () => 'foo',
  [OTHER_TEST_ACTION]: () => 'bar',
}, 'baz')

let state = 'test'
state = reducer(state, {type: TEST_ACTION})
console.log(state) // test (unexpected)
state = reducer(state, {type: OTHER_TEST_ACTION})
console.log(state) // bar (fine)
```

`TEST_ACTION` is broken, `OTHER_TEST_ACTION` is fine.

gaearon:
(Just FYI—while Redux supports symbols as action types, this is generally a bad idea. You won't be able to serialize your actions and replay them.)

billyjanitsch:
^ (Great point. I only point this out because Symbols are starting to be used elsewhere leading to bugs like [this one](https://github.com/fcomb/redux-logger/issues/19), and if Redux is going to support them then its utilities probably should as well.)

billyjanitsch:
Btw, looks like the problem is that Symbols aren't included in the array returned by `Object.keys` (by design -- see [this](http://blog.keithcirkel.co.uk/metaprogramming-in-es6-symbols/)). Should be a quick (if ugly) fix [here](https://github.com/acdlite/redux-actions/blob/master/src/handleActions.js#L5).

acdlite:
Good catch. I think `Object.getOwnPropertyNames()` would work?

gaearon:
@acdlite 

https://github.com/andreypopp/autobind-decorator/blob/954d825fcb585fa5747bba06505436cf56b3c287/src/index.js#L27-L38

acdlite:
Yuck, okay, haha.

acdlite:
Fixed in 0.8.0. Thanks!

billyjanitsch:
Thanks for the quick fix! :)

lukewestby:
Closing since this issue is fixed. Please reopen if anything else comes up. Thanks!

PR #113, opened 2016-07-31
combineActions: handle multiple actions with the same reducer

### LINKED ISSUE #31: Proposal: mechanism for handling multiple actions with the same reducer.
Hi all!

I'd like to propose a new feature to allow handling multiple actions with the same reducer in the case that someone has very general, over-arching properties of their state.

For example, when using redux on its own I've found it useful to have a `loading` property which gets reduced like this:

```javascript
function loading(state = false, action) {
  switch(action.type) {
  case LOAD_SOMETHING_START:
  case LOAD_SOMETHING_ELSE_START:
    return true;

  case LOAD_SOMETHING_COMPLETE:
  case LOAD_SOMETHING_ELSE_COMPLETE:
    return false;

  default:
    return state;
  }
}
```

I find this preferable to invoking an action to update `loading` within a composite/async action using redux-thunk because I would prefer not to have my state's other properties know about the `loading` property and keep that logic central to the reducer for `loading`.

Initially I'd like to propose being able to define an action handler with a name that contains a delimiter like `|`. To illustrate with the previous example:

```javascript
handleActions({
  'LOAD_SOMETHING_START | LOAD_SOMETHING_ELSE_START': () => true,
  'LOAD_SOMETHING_COMPLETE | LOAD_SOMETHING_ELSE_COMPLETE': () => false
}, false);
```

Alternatively, if it would be preferable not to expose the details about how to delimit action names, a function `combineActions()` might help:

```javascript
const startActions = combineActions('LOAD_SOMETHING_START', 'LOAD_SOMETHING_ELSE_START');
const completeActions = combineActions('LOAD_SOMETHING_COMPLETE', 'LOAD_SOMETHING_ELSE_COMPLETE');

handleActions({
  [startActions]: () => true,
  [completeActions]: () => false
}, false);
```

What do you think? I'd be thrilled to submit a PR for this if we agree it would be worthwhile.

yangmillstheory:
Close #31. Please see that issue for background and discussion.

### Changes

- adds `combineActions`, which takes a variadic list of action creators (functions), strings, or Symbols
- `handleAction`/`handleActions` can handle combined actions
- adds tests for the above

Note that there's a subtle change here that's backwards-compatible: 

> **Though Symbols are immutable, reducer checks against them are not done by reference, but by checking `toString` (as we do everything else).**

This won't break anyone, but it might surprise the few people that do use Symbol action types, [which is generally a bad idea](https://github.com/acdlite/redux-actions/issues/20#issuecomment-134340125).

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,9 +182,9 @@ const reducer = handleActions({
 
 ### `combineActions(...actionTypes)`
 
-Combine any number of action types or action creators.
+Combine any number of action types or action creators. `actionTypes` is a list of positional arguments which can be action type strings, symbols, or action creators.
 
-`actionTypes` is a list of positional arguments which can be action type strings, symbols, or action creators.
+This is useful for allowing your reducing multiple distinct actions with the same reducer.
 
 ```js
 const { increment, decrement } = createActions({
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ const reducer = handleActions({
 
 Combine any number of action types or action creators. `actionTypes` is a list of positional arguments which can be action type strings, symbols, or action creators.
 
-This is useful for allowing your reducing multiple distinct actions with the same reducer.
+This allows you to reduce multiple distinct actions with the same reducer.
 
 ```js
 const { increment, decrement } = createActions({
--- a/README.md
+++ b/README.md
@@ -96,19 +96,10 @@ This function exists because while action type strings can be joined with a conv
 const increment = createAction('INCREMENT', amount => ({ amount }))
 const decrement = createAction('DECREMENT', amount => ({ amount: -amount }))
 
-const reducer = handleAction(
-  combineActions(increment, decrement),
-  {
-    next(state, { payload: { amount } }) {
-      return { ...state, counter: state.counter + amount }
-    },
-
-    throw(state) {
-      return { ...state, counter: 0 }
-    },
-  },
-  { counter: 10 }
-)
+const reducer = handleAction(combineActions(increment, decrement), {
+  next: (state, { payload: { amount } }) => ({ ...state, counter: state.counter + amount }),
+  throw: state => ({ ...state, counter: 0 }),
+}, { counter: 10 })
 
 expect(reducer(undefined, increment(1)).to.deep.equal({ counter: 11 })
 expect(reducer(undefined, decrement(1)).to.deep.equal({ counter: 9 })
@@ -116,7 +107,7 @@ expect(reducer(undefined, increment(new Error)).to.deep.equal({ counter: 0 })
 expect(reducer(undefined, decrement(new Error)).to.deep.equal({ counter: 0 })
 ```
 
-This also works in when declaring reducers with the `next`/`throw` object format in `handleAction` and `handleActions`.
+This also works in when declaring multiple reducers with `handleActions`.
 
 ### `createActions(?actionsMap, ?...identityActions)`
 
--- a/README.md
+++ b/README.md
@@ -131,29 +131,6 @@ expect(actionThree(3)).to.deep.equal({
 });
 ```
 
-### `combineActions(...actionTypes)`
-
-Combine any number of action types or action creators.
-
-`actionTypes` is a list of positional arguments which can be action type strings, symbols, or action creators.
-
-```js
-const { increment, decrement } = createActions({
-  INCREMENT: amount => ({ amount }),
-  DECREMENT: amount => ({ amount: -amount }),
-})
-
-const reducer = handleAction(combineActions(increment, decrement), {
-  next: (state, { payload: { amount } }) => ({ ...state, counter: state.counter + amount }),
-  throw: state => ({ ...state, counter: 0 }),
-}, { counter: 10 })
-
-expect(reducer(undefined, increment(1)).to.deep.equal({ counter: 11 })
-expect(reducer(undefined, decrement(1)).to.deep.equal({ counter: 9 })
-expect(reducer(undefined, increment(new Error)).to.deep.equal({ counter: 0 })
-expect(reducer(undefined, decrement(new Error)).to.deep.equal({ counter: 0 })
-```
-
 ### `handleAction(type, reducer | reducerMap, ?defaultState)`
 
 ```js
@@ -201,6 +178,29 @@ const reducer = handleActions({
     counter: state.counter - action.payload
   })
 }, { counter: 0 });
+```
+
+### `combineActions(...actionTypes)`
+
+Combine any number of action types or action creators.
+
+`actionTypes` is a list of positional arguments which can be action type strings, symbols, or action creators.
+
+```js
+const { increment, decrement } = createActions({
+  INCREMENT: amount => ({ amount }),
+  DECREMENT: amount => ({ amount: -amount }),
+})
+
+const reducer = handleAction(combineActions(increment, decrement), {
+  next: (state, { payload: { amount } }) => ({ ...state, counter: state.counter + amount }),
+  throw: state => ({ ...state, counter: 0 }),
+}, { counter: 10 })
+
+expect(reducer(undefined, increment(1)).to.deep.equal({ counter: 11 })
+expect(reducer(undefined, decrement(1)).to.deep.equal({ counter: 9 })
+expect(reducer(undefined, increment(new Error)).to.deep.equal({ counter: 0 })
+expect(reducer(undefined, decrement(new Error)).to.deep.equal({ counter: 0 })
 ```
 
 ## Usage with middleware
--- a/README.md
+++ b/README.md
@@ -84,31 +84,6 @@ createAction('ADD_TODO')('Use Redux');
 
 `metaCreator` is an optional function that creates metadata for the payload. It receives the same arguments as the payload creator, but its result becomes the meta field of the resulting action. If `metaCreator` is undefined or not a function, the meta field is omitted.
 
-### `combineActions(...actionTypes)`
-
-Combine any number of action types or action creators.
-
-`actionTypes` is a list of positional arguments which can be action type strings or action creators.
-
-This function exists because while action type strings can be joined with a conventional delimiter, there is no obvious way for a library user to combine action creators.
-
-```js
-const { increment, decrement } = createActions({
-  INCREMENT: amount => ({ amount }),
-  DECREMENT: amount => ({ amount: -amount }),
-})
-
-const reducer = handleAction(combineActions(increment, decrement), {
-  next: (state, { payload: { amount } }) => ({ ...state, counter: state.counter + amount }),
-  throw: state => ({ ...state, counter: 0 }),
-}, { counter: 10 })
-
-expect(reducer(undefined, increment(1)).to.deep.equal({ counter: 11 })
-expect(reducer(undefined, decrement(1)).to.deep.equal({ counter: 9 })
-expect(reducer(undefined, increment(new Error)).to.deep.equal({ counter: 0 })
-expect(reducer(undefined, decrement(new Error)).to.deep.equal({ counter: 0 })
-```
-
 ### `createActions(?actionsMap, ?...identityActions)`
 
 ```js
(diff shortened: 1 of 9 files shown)


src/createAction.js
```diff
@@ -29,7 +29,7 @@ export default function createAction(type, payloadCreator, metaCreator) {
     return action;
   };
 
-  actionHandler.toString = () => type;
+  actionHandler.toString = () => type.toString();
```
yangmillstheory: This handles the case when a `Symbol` is passed as `type`.

yangmillstheory:
@timche can you take a look? 👌

src/createAction.js
timche: Do we really support Symbols even they are not recommended?

src/combineActions.js
```diff
@@ -0,0 +1,28 @@
+import isString from 'lodash/isString';
+import isFunction from 'lodash/isFunction';
+import isEmpty from 'lodash/isEmpty';
+import toString from 'lodash/toString';
+import isSymbol from 'lodash/isSymbol';
+
+export const ACTION_TYPE_DELIMITER = '|redux-action-delimiter|';
```
timche: What is this exactly for?

README.md
```diff
@@ -131,6 +131,29 @@ expect(actionThree(3)).to.deep.equal({
...
+const reducer = handleAction(combineActions(increment, decrement), {
+  next: (state, { payload: { amount } }) => ({ ...state, counter: state.counter + amount }),
+  throw: state => ({ ...state, counter: 0 }),
+}, { counter: 10 })
+
+expect(reducer(undefined, increment(1)).to.deep.equal({ counter: 11 })
+expect(reducer(undefined, decrement(1)).to.deep.equal({ counter: 9 })
+expect(reducer(undefined, increment(new Error)).to.deep.equal({ counter: 0 })
+expect(reducer(undefined, decrement(new Error)).to.deep.equal({ counter: 0 })
+```
```
timche: Could you please move this section under `handleActions`?

src/combineActions.js
yangmillstheory: It's used as a special character to join actions, then [split them again in the reducer](https://github.com/acdlite/redux-actions/pull/113/files#diff-b35552629f82c2c3dd19251a9cec2615R8).

This technique was [proposed by the reporter in the original issue](https://github.com/acdlite/redux-actions/issues/31#issue-107937867).

src/createAction.js
yangmillstheory: Yes, there's been a [test](https://github.com/acdlite/redux-actions/blob/master/src/__tests__/handleActions-test.js#L26) for this for quite a while (added in https://github.com/acdlite/redux-actions/commit/dd8ef00dd8883f8159d4de43f516ea52514b284e), and I didn't want to cause a "regression" with this change. I'm guessing it was fixed at the time because it easy and quick to fix.

Here's the original issue where the BDFL recommended against using it: https://github.com/acdlite/redux-actions/issues/20.

README.md
yangmillstheory: Good point. Makes sense to move it after all the code it references. f146fd6

src/createAction.js
timche: So the question is if we are going to drop the Symbols support in a future version like v1.0.0 to stick with Redux best practices.

README.md
```diff
@@ -180,6 +180,29 @@ const reducer = handleActions({
 }, { counter: 0 });
 ```
 
+### `combineActions(...actionTypes)`
+
+Combine any number of action types or action creators.
```
timche: I think the use case is missing in the description. For what is it good for?

src/createAction.js
yangmillstheory: I would be happy to make the change here. What do you think?

README.md
yangmillstheory: Ugh, true. Let me know if this works: 04bd910

README.md
```diff
@@ -180,6 +180,29 @@ const reducer = handleActions({
 }, { counter: 0 });
 ```
 
+### `combineActions(...actionTypes)`
+
+Combine any number of action types or action creators. `actionTypes` is a list of positional arguments which can be action type strings, symbols, or action creators.
+
+This is useful for allowing your reducing multiple distinct actions with the same reducer.
```
timche: I think there is a typo. `your` => `you`?

src/combineActions.js
timche: Is this somewhere visible like in logs? I thinking of shorten it a bit.

README.md
yangmillstheory: 😓  04197a5

src/combineActions.js
```diff
@@ -0,0 +1,28 @@
...
+
+  return { toString: () => combinedActionType };
+}
+
+function isValidActionTypes(actionTypes) {
+  if (isEmpty(actionTypes)) {
+    return false;
+  }
+  return actionTypes.every(isValidActionType);
+}
```
timche: For readability it's better imo to have this before `combineActions`.

src/combineActions.js
```diff
@@ -0,0 +1,28 @@
...
+
+function isValidActionType(actionType) {
+  return isString(actionType) || isFunction(actionType) || isSymbol(actionType);
+}
```
timche: For readability it's better imo to have this before `combineActions`.

src/combineActions.js
yangmillstheory: I haven't seen it using https://github.com/gaearon/redux-devtools for a few months in production. I don't think it'd come up using https://github.com/evgenyrodionov/redux-logger.

Actions are still dispatched per-type; it's just that registering them with a library reducer is different.

src/createAction.js
timche: As we are currently supporting them in `handleActions`, I would keep the support here as well.

src/combineActions.js
yangmillstheory: 1fb1eee

src/combineActions.js
yangmillstheory: 1fb1eee

src/combineActions.js
yangmillstheory: a8f3ffd

src/index.js
```diff
@@ -1,11 +1,13 @@
...
 import handleAction from './handleAction';
 import handleActions from './handleActions';
+import combineActions from './combineActions';
 import createActions from './createActions';
 
 export {
   createAction,
   createActions,
   handleAction,
+  combineActions,
```
timche: To keep a clean order, this could come after `handleActions`/last.

src/handleAction.js
```diff
@@ -1,19 +1,18 @@
...
+export default function handleAction(actionType, reducers, defaultState) {
+  const actionTypes = actionType.toString().split(ACTION_TYPE_DELIMITER);
 
   const [nextReducer, throwReducer] = isFunction(reducers)
     ? [reducers, reducers]
     : [reducers.next, reducers.throw].map(reducer => (isNil(reducer) ? identity : reducer));
 
   return (state = defaultState, action) => {
-    if (action.type !== typeValue) {
+    if (!includes(actionTypes, action.type.toString())) {
```
timche: We could use `Array.includes` here:

```js
actionTypes.includes(action.type.toString())
```

src/handleAction.js
```diff
@@ -1,19 +1,18 @@
...
-import isSymbol from 'lodash/isSymbol';
+import includes from 'lodash/includes';
+import { ACTION_TYPE_DELIMITER } from './combineActions';
 
-export default function handleAction(type, reducers, defaultState) {
-  const typeValue = isSymbol(type)
-    ? type
-    : type.toString();
+export default function handleAction(actionType, reducers, defaultState) {
+  const actionTypes = actionType.toString().split(ACTION_TYPE_DELIMITER);
```
timche: If `actionType` is a plain string, does `actionType.toString()` really work?

timche: Nvm, just tried it out, it works, just didn't know that, because it looks weird :D

src/index.js
yangmillstheory: Ugh, sorry, I should have been catching these. It's been a while since the PR, and I was on vacation 😓 

5dd2f29

src/handleAction.js
timche: For performance we could do a check before this if `actionType` includes `ACTION_TYPE_DELIMITER`, otherwise skip this transformation, because I think the majority of `actionTypes` will be non-combined action types.

src/handleAction.js
yangmillstheory: d87fef6

timche: Related to my [comment above](https://github.com/acdlite/redux-actions/pull/113/files#r77036110), this should do either the old check `action.type !== typeValue` or this if it is a combined action.

src/handleAction.js
yangmillstheory: Hm, I'm not sure. I think the code is a bit clearer without the special cases, and we don't know that performance will be an issue yet here (and I haven't seen it yet in a production app). Can we optimize when we know we'll need it and leave the code without special cases for now (for better readability)?

src/combineActions.js
timche: You are right, it's good as identifier.

src/handleAction.js
```diff
@@ -1,19 +1,17 @@
...
 
   const [nextReducer, throwReducer] = isFunction(reducers)
     ? [reducers, reducers]
     : [reducers.next, reducers.throw].map(reducer => (isNil(reducer) ? identity : reducer));
 
   return (state = defaultState, action) => {
-    if (action.type !== typeValue) {
+    if (!actionTypes.includes(action.type.toString())) {
```
timche: If we are going with [my performance comment above](https://github.com/acdlite/redux-actions/pull/113/files#r77036110), then this has to either do the check `action.type !== typeValue` like before or this.

src/handleAction.js
timche: The special case for me is `combineActions`. My concern was only that we are doing `.toString()` and `.split()` for every action that will be received in the reducer. But tbh, I don't know if there will be performance issues neither, was just an assumption.

We could write a benchmark to find out.

src/handleAction.js
yangmillstheory: Wait, now I remember why I didn't use native `includes`. It breaks on older versions of node https://travis-ci.org/acdlite/redux-actions/builds/156589380.

src/handleAction.js
timche: Here is another solution:

We could leave this line like it was before and change only L14 to: 

```js
if (!actionType.include(action.type) {
```

We make use of [`String.prototype.includes()`](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/includes) and we have to change less. Performance wise it would be less overhead imo.

Btw, why I `.toString()` needed when calling `action.type`?

src/handleAction.js
yangmillstheory: I reverted it for now: e2fbec34

timche: You are right, we would need `babel-polyfill` for that to make it run.

src/handleAction.js
timche: This will also fail and needs `babel-polyfill`. Instead this would also work with `lodash/includes`:

```js
if (!includes(actionType, action.type) {
```

timche: Or simply [`String.prototype.indexOf()`](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf):

```js
if (actionType.indexOf(action.type) === -1) {
```

yangmillstheory: It's used to keep compatibility with Symbol support; otherwise [this test](https://github.com/acdlite/redux-actions/blob/812d5f208f941c181b716bdb9fe524235ae51e96/src/__tests__/handleActions-test.js#L26) and another Symbol-related test fails.

timche: Yes, that's why this line could be reverted to keep the compatibility and only the if part could change or am I missing something?

src/combineActions.js
timche:
```diff
 import toString from 'lodash/toString';
 import isSymbol from 'lodash/isSymbol';

-export const ACTION_TYPE_DELIMITER = '|redux-action-delimiter|';
+export const ACTION_TYPE_DELIMITER = ' || ';

 function isValidActionType(actionType) {
   return isString(actionType) || isFunction(actionType) || isSymbol(actionType);
```

yangmillstheory: LGTM

timche:
The PR looks good to me now!

**P.S. Don't forget to squash before merge ;)**

yangmillstheory:
Thanks for the review and time 👍

Squashing... 🔨  :boom:

yangmillstheory:
Wait, should we ask @acdlite to enable only squash merging via the settings? Seems like it would be useful to automatically enforce this process.

timche:
Sure, I will ping him.

Issue #39, opened 2015-10-17
createActions


appsforartists:
Just as there's a `handleActions`, perhaps there should be a `createActions` as sugar for creating multiple actions?

```javascript
createActions(
  {
    [type1]:  payload1Creator,

    [type2]:  {
                "payload":  payload2Creator,
                "meta":     meta2Creator,
              }
  }
);
```

This seems a bit nicer (and less redundant) than

```javascript
{
  [type1]: createAction(type1, payload1Creator),
  [type2]: createAction(type2, payload2Creator, meta2Creator),
}
```

msafi:
Came here to ask for this.

In my case I did

```
mapValues(myActionTypes, (actionType) => createAction(actionType));
```

appsforartists:
@msafi https://github.com/pauldijou/redux-act looks pretty sweet too.

timche:
Since there are no more opinions on that, I'll close it for now. Re-open if necessary.

PR #114, opened 2016-07-31
createActions: create multiple action creators at once

yangmillstheory:
### Changes

- `createActions` API method with tests

Please see #104, #39 , #83 for background and discussion, and read the new [API documentation](https://github.com/acdlite/redux-actions/blob/feat/create-actions/README.md#createactionsactionsmap-actiontypes).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,6 +16,6 @@
     "no-param-reassign": 0,
     "no-reserved-keys": 0,
     "space-infix-ops": 0,
-    "array-bracket-spacing": [2, "never"],
+    "array-bracket-spacing": [2, "never"]
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,6 +15,8 @@
     "no-unused-expressions": 0,
     "no-param-reassign": 0,
     "no-reserved-keys": 0,
-    "space-infix-ops": 0
+    "space-infix-ops": 0,
+    array-bracket-spacing: [2, "never"],
+    object-curly-spacing: [2, "never"]
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,6 +17,5 @@
     "no-reserved-keys": 0,
     "space-infix-ops": 0,
     "array-bracket-spacing": [2, "never"],
-    "object-curly-spacing": [2, "always"]
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,7 +16,7 @@
     "no-param-reassign": 0,
     "no-reserved-keys": 0,
     "space-infix-ops": 0,
-    array-bracket-spacing: [2, "never"],
-    object-curly-spacing: [2, "never"]
+    "array-bracket-spacing": [2, "never"],
+    "object-curly-spacing": [2, "always"]
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,7 +15,6 @@
     "no-unused-expressions": 0,
     "no-param-reassign": 0,
     "no-reserved-keys": 0,
-    "space-infix-ops": 0,
-    "array-bracket-spacing": [2, "never"]
+    "space-infix-ops": 0
   }
 }
--- a/README.md
+++ b/README.md
@@ -93,14 +93,12 @@ Returns an object mapping action types to action creators. The keys of this obje
 ```js
 const { actionOne, actionTwo, actionThree } = createActions({
   // function form; payload creator defined inline
-  ACTION_ONE(key, value) {
-    return { [key]: value };
-  },
+  ACTION_ONE: (key, value) => ({ [key]: value }),
   
   // array form
   ACTION_TWO: [
-    (first, second) => [first, second],  // payload
-    (first, second) => { first, second } // meta
+    (first) => first,               // payload
+    (first, second) => ({ second }) // meta
   ],
 
   // trailing action type string form; payload creator is the identity
@@ -113,8 +111,8 @@ expect(actionOne('key', 1)).to.deep.equal({
 
 expect(actionTwo('first', 'second')).to.deep.equal({
   type: 'ACTION_TWO',
-  payload: ['first', 'second'],
-  meta: { first: 'first', second: 'second' }
+  payload: ['first'],
+  meta: { second: 'second' }
 });
 
 expect(actionThree(3)).to.deep.equal({
--- a/README.md
+++ b/README.md
@@ -78,17 +78,17 @@ createAction('ADD_TODO')('Use Redux');
 
 `metaCreator` is an optional function that creates metadata for the payload. It receives the same arguments as the payload creator, but its result becomes the meta field of the resulting action. If `metaCreator` is undefined or not a function, the meta field is omitted.
 
-### `createActions(?actionsMap, ?...actionTypes)`
+### `createActions(?actionsMap, ?...identityActions)`
 
-Returns an object mapping action types to action creators. The keys of this object are camel-cased from the keys in `actionsMap` and the values of the `actionTypes` arguments; the values are the action creators.
+Returns an object mapping action types to action creators. The keys of this object are camel-cased from the keys in `actionsMap` and the string literals of `identityActions`; the values are the action creators.
 
-`actionsMap` is an optional object with action types as keys, and whose values can be
+`actionsMap` is an optional object with action types as keys, and whose values **must** be either
 
 - a function, which is the payload creator for that action
 - an array with `payload` and `meta` functions in that order, as in [`createAction`](#createactiontype-payloadcreator--identity-metacreator)
     - `meta` is **required** in this case (otherwise use the function form above)
 
-`actionTypes` is an optional list of positional arguments that are action type strings; these action types will use the identity payload creator.
+`identityActions` is an optional list of positional string arguments that are action type strings; these action types will use the identity payload creator.
 
 ```js
 const { actionOne, actionTwo, actionThree } = createActions({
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -78,6 +78,51 @@ createAction('ADD_TODO')('Use Redux');
...
 
+##### `createActions(?actionsMap, ?...actionTypes)`
+
+Returns an object mapping action types to action creators. The keys of this object are camel-cased from the keys in `actionsMap`; the values are the action creators.
+
+`actionsMap` is an optional object with action types as keys, and whose values can be
+
+- a function, which is the payload creator for that action
+- an object with `payload` and `meta` keys, mapping to payload and meta creators
+    - `meta` is **required** in this case to distinguish it meaningfully from the function form above
```
timche: `form` => `from`

timche: I don't understand what you mean with `to distinguish it meaningfully from the function from above`. Why do we need to distinguish them? Isn't the key enough as identifier?

README.md
```diff
@@ -78,6 +78,51 @@ createAction('ADD_TODO')('Use Redux');
...
+```js
+const { actionOne, actionTwo, actionThree } = createActions({
+  ACTION_ONE(key, value) {
+    return { [key]: value };
+  },
+
+  ACTION_TWO: {
+    payload(first, second) {
+      return [ first, second ];
+    }
```
timche: Comma is missing: `},`

README.md
yangmillstheory: Ah, I mean:

> If you're using the object form but not supplying the `meta` key, you may as well be using the single `function` form

Does that make sense? Can I word it better or drop the restriction?

README.md
timche: I'd suggest to use an arrow function, so they will become one liners.

```js
payload: (first, second) => [first, second],
meta: (first, second) => ({ first , second })
```

Also no spaces at the brackets please ;)

README.md
```diff
@@ -78,6 +78,51 @@ createAction('ADD_TODO')('Use Redux');
...
+    meta(first, second) {
+      return { first, second };
+    }
```
timche: I'd suggest to use arrow functions, so they will become one liners.

```js
payload: (first, second) => [first, second],
meta: (first, second) => ({ first , second })
```

Also no spaces at the brackets please ;)

yangmillstheory: 4f6be56

I feel like we should lint for that - we're doing linting for this at my job.

README.md
yangmillstheory: 4f6be56 thanks

README.md
```diff
@@ -78,6 +78,47 @@ createAction('ADD_TODO')('Use Redux');
 
 `metaCreator` is an optional function that creates metadata for the payload. It receives the same arguments as the payload creator, but its result becomes the meta field of the resulting action. If `metaCreator` is undefined or not a function, the meta field is omitted.
 
+##### `createActions(?actionsMap, ?...actionTypes)`
```
timche: Should be `###`, right?

yangmillstheory: Yep. 8388ad4

README.md
yangmillstheory: OK, so I turned this on, and it auto-fixed a bunch of stuff: 99e95ba

README.md
timche: Ah, I understand now. It was indeed a bit confusing.

The only problem I now see is, it is not really self-explaining if you see the function form and the object form together for the first time, but they are basically the same, except that the object form has additionally `meta`.

How about inheriting the function form from `createAction` and write `actionMap` like this:

```js
ACTION_ONE: [
  (first, second) => [first, second],
  (first, second) => {first, second]
]
```

The first value in the array is `payload` and second `meta`.

Just an idea to reduce the possibilities how we can write the `actionMap`.

yangmillstheory: I like that approach, but how do you think it compares with the clarity of the object form? 

With the array approach, I have to know that the positional indices `0`,`1`, map to `payload` and `meta`, whereas in the object form, I don't have to rely on that convention, since it's named.

timche: I was just adopted the ordering from `createAction` where it is `(type, payloadCreator, metaCreator)`.

yangmillstheory: I can do this 👍 . Should I still leave support for the object form?

timche: I would drop it. Hope you are really fine with this :P

yangmillstheory: Haha. I like the object approach better, but I also agree with your reasons: less verbose and matches the order of `createAction`.

If people request it we can add it back.

yangmillstheory: So let's clarify the requirements for the array form if one function is passed. Should we:

- throw an error?
- interpret the one function as `payload`
- interpret the one function as `meta`, and default `payload` to `identity`?

timche: > throw an error?
Yeah, why not.

> interpret the one function as payload
Yes.

> interpret the one function as meta, and default payload to identity
Yes.

What you can basically do is, just take the array in spread it into the `createActions`: `createAction(key, [...value])`

.eslintrc
```diff
@@ -15,6 +15,8 @@
     "no-unused-expressions": 0,
     "no-param-reassign": 0,
     "no-reserved-keys": 0,
-    "space-infix-ops": 0
+    "space-infix-ops": 0,
+    array-bracket-spacing: [2, "never"],
+    object-curly-spacing: [2, "never"]
```
timche: Nooo, I didn't mean to remove the spaces in curly braces `{}` as well, sorry 😅 Only in brackets `[]`. Normally the [`airbnb`-preset](https://github.com/airbnb/javascript/search?utf8=%E2%9C%93&q=array-bracket-spacing) should throw an error already.

README.md
yangmillstheory: Wait, those three bullets were mutually exclusive 😄.

>What you can do now, just take the array and spread it into `createActions`: `createAction(key, ...value)`

I think this fourth option makes the most sense, since we're using the signature of `createAction`.

.eslintrc
yangmillstheory: Removed in 558de5e.

yangmillstheory:
@timche can you take another look?

README.md
```diff
@@ -78,6 +78,51 @@ createAction('ADD_TODO')('Use Redux');
...
+    - `meta` is **required** in this case (otherwise use the function form above)
+
+`actionTypes` is an optional list of positional arguments that are action type strings; these action types will use the identity payload creator.
+
+```js
+const { actionOne, actionTwo, actionThree } = createActions({
+  // function form; payload creator defined inline
+  ACTION_ONE(key, value) {
+    return { [key]: value };
+  },
```
timche: Can we use an arrow function here please?

README.md
```diff
@@ -78,6 +78,51 @@ createAction('ADD_TODO')('Use Redux');
...
+  
+  // array form
+  ACTION_TWO: [
+    (first, second) => [first, second],  // payload
+    (first, second) => { first, second } // meta
```
timche: `{ first, second }` => `({ first, second })`

I would add a third argument or only take the second, just for clarification how this works, so beginners understand directly, that `payload` and `meta` get the same arguments.

Example:

```js
(first) => first,
(first, second) => ({ second })
```

.eslintrc
```diff
@@ -15,6 +15,7 @@
     "no-unused-expressions": 0,
     "no-param-reassign": 0,
     "no-reserved-keys": 0,
-    "space-infix-ops": 0
+    "space-infix-ops": 0,
+    "array-bracket-spacing": [2, "never"]
```
timche: I'm confused, doesn't [airbnb has this rule already](https://github.com/airbnb/javascript/blob/a533e2dd35ebc05e246698e65a6e23cbd342c240/packages/eslint-config-airbnb-base/rules/style.js#L4)?

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "npm run clean && babel src --out-dir lib --ignore *-test.js",
     "clean": "rimraf lib",
-    "lint": "esw src --color",
+    "lint": "esw src --color --fix",
```
timche: Will you remove this later or can we separate this into another command called `lint:fix`? We need this command without auto-fix for CI.

src/__tests__/createActions-test.js
```diff
@@ -0,0 +1,188 @@
...
+    expect(actionTwo('value', 2)).to.deep.equal({
+      type: 'ACTION_TWO',
+      payload: ['value', 2]
+    });
+  });
+
+  it('should use the identity payload creator if the reducer value is undefined', () => {
+    const { action1, action2 } = createActions({
+      ACTION_1: undefined,
+      ACTION_2: undefined
```
timche: Do we really want to allow this and not forcing it to string form only?

timche:
Only small things, otherwise really neat implementation 👌🏻

package.json
yangmillstheory: 24b9c690

README.md
yangmillstheory: 6964521f

README.md
yangmillstheory: 6964521f

.eslintrc
yangmillstheory: Yes - thanks for that. I put this in from [this comment](https://github.com/acdlite/redux-actions/pull/114#discussion_r73225919), thinking it was off, but obviously markdown isn't linted. 👍

yangmillstheory: 569fa7d1

src/__tests__/createActions-test.js
yangmillstheory: 998cd0ea

yangmillstheory:
@timche I have another round of revisions. Let me know if there's anything else. Thanks 👌 .

timche:
LGTM.

yangmillstheory:
Ugh forgot to squash dcc7026 in; used to having only squash merges allowed in the settings.

I stripped dcc7026 and squashed with 812d5f20.

PR #115, opened 2016-08-01
Only omit payload if undefined

jrasky:
Include a `null` payload if defined

### DIFF
--- a/src/createAction.js
+++ b/src/createAction.js
@@ -13,7 +13,7 @@ export default function createAction(type, payloadCreator, metaCreator) {
     };
 
     const payload = hasError ? args[0] : finalPayloadCreator(...args);
-    if (!(payload === null || payload === undefined)) {
+    if (payload !== undefined) {
       action.payload = payload;
     }
 
--- a/src/createAction.js
+++ b/src/createAction.js
@@ -1,4 +1,5 @@
 import identity from 'lodash/identity';
+import isUndefined from 'lodash/isUndefined';
 
 export default function createAction(type, payloadCreator, metaCreator) {
   const finalPayloadCreator = typeof payloadCreator === 'function'
@@ -13,7 +14,7 @@ export default function createAction(type, payloadCreator, metaCreator) {
     };
 
     const payload = hasError ? args[0] : finalPayloadCreator(...args);
-    if (payload !== undefined) {
+    if (!isUndefined(payload)) {
       action.payload = payload;
     }
 
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -95,11 +95,12 @@ describe('createAction()', () => {
 
       const explictNullAction = createAction(type)(null);
       expect(explictNullAction).to.deep.equal({
-        type
+        type,
+        payload: null
       });
 
       const baz = '1';
-      const actionCreator = createAction(type, null, () => ({ bar: baz }));
+      const actionCreator = createAction(type, undefined, () => ({ bar: baz }));
       expect(actionCreator()).to.deep.equal({
         type,
         meta: {


yangmillstheory:
Thanks for the contribution. I like not distinguishing between `null` and `undefined`, but have you seen https://github.com/acdlite/redux-actions/pull/48, and in particular https://github.com/acdlite/redux-actions/pull/48#issuecomment-220525042?

I think @timche might have an opinion about this change.

jrasky:
The ECMAScript spec distinguishes between `undefined` as "no assigned value" and `null` as "intentionally absent value." I

yangmillstheory:
ping @timche 

I'm happy to make an issue to not conflate `undefined` and `null` going forward.

timche:
This is a hard one imo. For example [react-redux](https://github.com/reactjs/react-redux/blob/master/docs/api.md) is using `null` in their `connect` function to omit arguments.

I also prefer `null` over `undefined`.

@jrasky can you please describe the problem you see with it with an example?

jrasky:
I don't doubt there's cases where treating `null` and `undefined` the same way is fine, and react-redux shows such an example. I do think, however, that the ECMAScript standard is a higher source of truth than react-redux.

The case I'm running into is where I'm using `undefined` to mean "no data yet," and `null` to mean "no data available," which is roughly in line with how the spec defines them.

Very specifically, I don't expect that passing in `null` as a payload will result in the reducer receiving no payload, since `null` in javascript means "explicitly no value," unlike `undefined`.

There are always "more code" approaches to solving nearly any problem, my concern here is with having more correct, spec-compliant code.

timche:
For me personally, the open source/JS community is a higher source than the ECMAScript standard, because it represents how JS is actually used widely.

I'd be happy if you could provide some links.

I'm willing to change it, but that would also mean, it is a breaking change and we have to release a major version.

jrasky:
To be clear, I think your usage here is different from how redux-react uses null, because in the redux-actions case the value is "passed along."

There's a few thinks I can link you to in the standard:
* [The section on `undefined`, with the section on `null` shortly thereafter](http://www.ecma-international.org/ecma-262/6.0/#sec-undefined-value)
* `undefined` is used to omit arguments in several places like, [here](http://www.ecma-international.org/ecma-262/6.0/#sec-promise.prototype.catch), [here](http://www.ecma-international.org/ecma-262/6.0/#sec-createglobalvarbinding), and [here](http://www.ecma-international.org/ecma-262/6.0/#sec-ecmascript-function-objects-call-thisargument-argumentslist).

redux itself [treats `null` and `undefined` differently](https://github.com/reactjs/redux/blob/master/src/createStore.js#L181), although in a somewhat implicit fashion.

It is a somewhat significant change, but also one that's not very difficult to work around. That being said, the codebase I'm working on and the standard both seem to indicate that `undefined` is the "omitted argument" value.

timche:
I'd like to fix this with v1.0.0, also to follow [semver](http://semver.org/). Everybody's fine with that?

src/createAction.js
```diff
@@ -13,7 +13,7 @@ export default function createAction(type, payloadCreator, metaCreator) {
     };
 
     const payload = hasError ? args[0] : finalPayloadCreator(...args);
-    if (!(payload === null || payload === undefined)) {
+    if (payload !== undefined) {
```
timche: Could you please use the lodash function `lodash/isUndefined` here?

jrasky: Sure, but just consider that it does exactly the same thing: https://github.com/lodash/lodash/blob/master/lodash.js#L11878

TylerBrock:
Lets get this merged! We just bumped into a situation where our payload is null (explicitly) and the current behavior was surprising.

timche:
@jrasky Can please have a look at [this](https://github.com/jrasky/redux-actions/blob/a12e9ee91bfa1ed85c6b331a3d1dba48a68eb955/src/__tests__/createAction-test.js#L38-L47)?

This looks a bit confusing to me. Are we now supposed to use `null` or `undefined` there?

timche:
Continuing in #128 to speed up the process.

PR #117, opened 2016-08-15
add UMD build process

timche:


README.md
```diff
@@ -1,20 +1,26 @@
...
+## Installation
+
+```bash
 npm install --save redux-actions
 ```
-```js
-import { createAction, handleAction, handleActions } from 'redux-actions';
-```
+
+If you don’t use npm, you may grab the latest [UMD](https://npmcdn.com/redux-actions@latest/dist) build from [npmcdn](https://npmcdn.com) (either a [development](https://npmcdn.com/redux-actions@latest/dist/redux-actions.js) or a [production](https://npmcdn.com/redux-actions@latest/dist/redux-actions.min.js) build). The UMD build exports a global called `window.ReduxActions` if you add it to your page via a `<script>` tag. We *don’t* recommend UMD builds for any serious application, as most of the libraries complementary to Redux are only available on [npm](http://npmjs.org).
```
yangmillstheory: Can we **bold**, CAPS, [linkify](www.npmjs.com), or monospace `npm`?

What are some libraries complementary to Redux that are only available on NPM? Would it be worth listing some?

webpack.config.babel.js
```diff
@@ -0,0 +1,48 @@
...
+      }
+    })
+  );
+}
+
+export default {
+  entry: path.join(__dirname, 'src/index.js'),
+  output: {
+    path: path.join(__dirname, 'dist'),
+    filename,
```
yangmillstheory: What do you think about inlining `filename`? It's not used elsewhere, and when I got down here I had to look for its definition at the beginning of the file.

README.md
yangmillstheory: Also, when will the files on the CDN exist? When they are published?

package.json
```diff
@@ -5,15 +5,19 @@
...
+    "lint:fix": "npm run lint -- --fix",
     "lint:watch": "npm run lint -- --watch",
     "prepublish": "npm run lint && npm run test && npm run build",
     "test": "mocha --compilers js:babel-register src/**/*-test.js",
     "test:watch": "npm run test -- --watch src/**/*-test.js"
   },
   "files": [
-    "lib"
+    "lib",
+    "dist"
```
yangmillstheory: Are `build:umd` and `build:umd:min` part of the Travis build? If not, there will be no `dist/` to include, no?

webpack.config.babel.js
```diff
@@ -0,0 +1,48 @@
...
+const filename = `redux-actions${production ? '.min' : ''}.js`;
+
+const plugins = [
+  new webpack.optimize.OccurenceOrderPlugin(),
+  new webpack.DefinePlugin({
+    'process.env.NODE_ENV': JSON.stringify(NODE_ENV)
+  })
+];
+
+if (production) {
```
yangmillstheory: What do you think of `NODE_ENV === 'production'` inline here, since `production` is not used elsewhere and the triple-equals is already pretty readable?

README.md
timche: > Can we bold, CAPS, linkify, or monospace npm?

The second `npm` is linkified, I will do it for the first one as well, good catch.

> What are some libraries complementary to Redux that are only available on NPM? Would it be worth listing some?

Well, there are [tons of packages](https://www.npmjs.com/search?q=redux) 🙈 I've just adopted from [react-redux](https://github.com/reactjs/react-redux#installation).

> Also, when will the files on the CDN exist? When they are published?

They are available on npmcdn as soon as we are publishing this to npm. There is no additional effort for us.

package.json
timche: Forgot to add it to `prepublish`.

webpack.config.babel.js
timche: It is also used in `filename`.

timche:
@yangmillstheory ready for another review!

yangmillstheory:
LGTM. Thanks!

PR #119, opened 2016-08-20
Depend on specific lodash modules instead of all of lodash.

jirwin:
Instead of depending on all of lodash, we can only depend on the specific functions that are used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,15 @@
     "webpack": "^1.13.1"
   },
   "dependencies": {
-    "lodash": "^4.13.1",
+    "lodash.camelcase": "^4.3.0",
+    "lodash.identity": "^3.0.0",
+    "lodash.isarray": "^4.0.0",
+    "lodash.isfunction": "^3.0.8",
+    "lodash.isnil": "^4.0.0",
+    "lodash.isplainobject": "^4.0.6",
+    "lodash.isstring": "^4.0.1",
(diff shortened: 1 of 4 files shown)


yangmillstheory:
Thanks for the contribution.

Pulling in all of `lodash` was an intentional design choice; see https://github.com/acdlite/redux-actions/pull/109#discussion_r71062856.

@timche; thoughts?

timche:
Thanks for pointing out. But since we are using ES6 `import` and soon the tree-shaking ability e.g. in [Webpack 2](https://github.com/webpack/webpack), we are already importing the specific lodash functions with the slash notation alias the relative path. In the end, it won't make any difference but more maintaining effort for all the specific libraries we use. Also the ability to have access to the full lodash library makes it easier for pull requests to make use of them without the need to add them as new dependency to the `package.json`. Also the article "[The cost of small modules](https://nolanlawson.com/2016/08/15/the-cost-of-small-modules/)" is a great article about this topic.

jirwin:
@yangmillstheory Aha. I was looking for a comment like that before pushing my PR, but didn't have luck finding it. 

`loadash` on its own is just a large dependency(file count wise), so adding it to a project where dependencies are committed was a bit shocking. The 1000+ files did not make Github very happy.

Since it is an intentional choice, I'll go ahead and close this. Thanks!

Issue #108, opened 2016-07-13
`defaultState` not passed to `handleAction` from  `handleActions`


vanthiyathevan:
Shouldn't it be passed to `handleAction` from `handleActions`?

yangmillstheory:
https://github.com/acdlite/redux-actions#handleactionsreducermap-defaultstate

I don't think this would work.

The `defaultState` parameter is passed to the combined (reduced) reducer. 

Since we're using [reduce-reducers](https://github.com/acdlite/reduce-reducers) which runs the reducers serially, it will actually be passed to the first reducer in the chain in case the state passed in was `undefined`.

Reducers need to return a new state object (not `undefined`):

> (previousState, action) => newState

That's from the [Redux docs](http://redux.js.org/docs/basics/Reducers.html). Therefore, there will never be a need to pass the same `defaultState` to remaining reducers in the chain.

PR #127, opened 2016-09-05
Make defaultState required when creating reducers

### LINKED ISSUE #18: Make defaultState non-optional
Since we have to return a non-undefined state in redux, we should probably make the `defaultState` parameter for `handleActions` required. Thoughts?

yangmillstheory:
Close #18.

See that issue and also https://github.com/reactjs/redux/issues/514 for background and discussion. 
***
**TL;DR**

We can prevent downstream reducer sanity errors [like this one](https://github.com/reactjs/redux/blob/master/src/combineReducers.js#L63) by throwing earlier.

### DIFF
--- a/src/assertDefaultState.js
+++ b/src/assertDefaultState.js
@@ -1,0 +1,7 @@
+import isUndefined from 'lodash/isUndefined';
+
+export default function assertDefaultState(defaultState, actionTypes) {
+  if (isUndefined(defaultState)) {
+    throw new Error(`Expected defaultState for reducer handling ${actionTypes.join(', ')} to be defined`)
+  }
+}
--- a/src/assertDefaultState.js
+++ b/src/assertDefaultState.js
@@ -2,6 +2,8 @@ import isUndefined from 'lodash/isUndefined';
 
 export default function assertDefaultState(defaultState, actionTypes) {
   if (isUndefined(defaultState)) {
-    throw new Error(`Expected defaultState for reducer handling ${actionTypes.join(', ')} to be defined`)
+    throw new Error(
+      `Expected defaultState for reducer handling ${actionTypes.join(', ')} to be defined`
+    );
   }
 }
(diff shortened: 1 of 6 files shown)


timche:
We are explicitly requiring it, so let's note it as breaking change, regardless of how Redux is handling it.

src/assertDefaultState.js
```diff
@@ -0,0 +1,9 @@
+import isUndefined from 'lodash/isUndefined';
+
+export default function assertDefaultState(defaultState, actionTypes) {
+  if (isUndefined(defaultState)) {
+    throw new Error(
+      `Expected defaultState for reducer handling ${actionTypes.join(', ')} to be defined`
+    );
+  }
+}
```
timche: I'd suggest to move this function to a folder named `utils` and name that file `assertions.js` or sth like this, where we can write more functions like this in a single file.

timche: Generally that's a good idea to extract the assertions 👍

timche:
Btw, I've changed the base branch to `v1.0.0`, so we can merge some other stuff into `master` that's not supposed to be in `v1.0.0`.

src/assertDefaultState.js
timche: Also, we will rewrite this later with `invariant`, correct?

yangmillstheory: Yes, or we may as well do it now.

yangmillstheory:
I've updated the PR; it's been a while. We're now using `invariant` for runtime checks since https://github.com/acdlite/redux-actions/pull/141.

@JaKXz @timche can you take another look?

src/handleActions.js
```diff
@@ -3,8 +3,13 @@ import ownKeys from './ownKeys';
...
 
 export default function handleActions(handlers, defaultState) {
-  const reducers = ownKeys(handlers).map(type => handleAction(type, handlers[type]));
+  const reducers = ownKeys(handlers).map(type =>
+    handleAction(
+      type,
+      handlers[type],
+      defaultState
+    )
+  );
```
yangmillstheory: Relevant: https://github.com/acdlite/redux-actions/issues/108

Issue #122, opened 2016-08-31
v1.0.0 Roadmap


timche:
With v1.0.0 we are releasing our first version that comes with breaking changes:

## Breaking Changes:
- [x] #115: Only omit payload if undefined
- [ ] #127: Make defaultState required when creating reducers
- [x] #145: Fail early for non-FSA actions

timche:
I would also like to introduce [Flowtype](https://flowtype.org/) with this version as it is getting popular and already being used in many React application as well as in [create-react-app](https://github.com/facebookincubator/create-react-app).

cc @yangmillstheory

yangmillstheory:
Sounds good.

I remember you had a great idea to clean up the documentation too. Just wanted to make sure that's not forgotten (not sure which epic it goes in).

timche:
Oh, yeah, documentation cleanup/improvement would also be nice. I will add it.

timche:
Because #115 seems to be more relevant than expected, we can release that as only change in v1.0.0 and everything else after to speed up the release.

cc @yangmillstheory

yangmillstheory:
That sounds good. Are you going to take #115?

timche:
Yes @yangmillstheory

yangmillstheory:
@timche can I take care of #47 in v1.0.0?

timche:
That would be huge breaking change. Could be sth for v2.

yangmillstheory:
What about #18? 

I'm not sure if this would also be breaking actually, since I think Redux throws errors on missing default state.

#18 just ensures the errors are thrown earlier.

timche:
Yes, this is actually breaking, since we are making this required, even it was throwing an error before.

yangmillstheory:
OK. Please see #127 😄 .

yangmillstheory:
Can we add type-checking of `payloadCreator` and `metaCreator` to this?

See the first example in https://github.com/acdlite/redux-actions/pull/128#issuecomment-244896176.

yangmillstheory:
I've ported this roadmap issue over to https://github.com/acdlite/redux-actions/projects/1.

PR #128, opened 2016-09-06
Omit payload on undefined only

timche:
Continuing #115 and part of #122 (v1.0.0).

### DIFF
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -95,7 +95,8 @@ describe('createAction()', () => {
 
       const explictNullAction = createAction(type)(null);
       expect(explictNullAction).to.deep.equal({
-        type
+        type,
+        payload: null
       });
 
       const baz = '1';
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -93,12 +93,6 @@ describe('createAction()', () => {
         type
       });
 
-      const explictNullAction = createAction(type)(null);
-      expect(explictNullAction).to.deep.equal({
-        type,
-        payload: null
-      });
-
       const baz = '1';
       const actionCreator = createAction(type, null, () => ({ bar: baz }));
       expect(actionCreator()).to.deep.equal({
--- a/src/createAction.js
+++ b/src/createAction.js
@@ -1,4 +1,5 @@
 import identity from 'lodash/identity';
+import isUndefined from 'lodash/isUndefined';
 
 export default function createAction(type, payloadCreator, metaCreator) {
   const finalPayloadCreator = typeof payloadCreator === 'function'
@@ -13,7 +14,7 @@ export default function createAction(type, payloadCreator, metaCreator) {
     };
 
     const payload = hasError ? args[0] : finalPayloadCreator(...args);
-    if (!(payload === null || payload === undefined)) {
+    if (!isUndefined(payload)) {
       action.payload = payload;
     }
 


timche:
@yangmillstheory This PR is slightly different than #115. Please have a look at the changes. You'll see that I've only changed a single line to fix the tests. It now looks a bit weird to me, because at some parts in `createAction` we still have `null` (for example [here](https://github.com/acdlite/redux-actions/blob/e875b471aa2b1ed9262dc823ccdb7f9e876d5e22/src/__tests__/createAction-test.js#L73)).

So in the end, only the identity function is affected by this change. Do you think this is overall still correct? 

I actually thought that we are using `undefined` everywhere to explicitly omitting the payload.

Examples:
```js
// Omit payload with undefined to use identity function and to define meta
const actionCreatorA = createAction(type, undefined, () => 'foo');

// Omit payload in identity function
const actionCreatorB= createAction(type);
actionCreatorB(undefined)
```

timche:
Ping @jrasky, @TylerBrock, @noahsilas, @vbud, @arb.

jrasky:
Hmm, your list confuses me a little bit. This PR is such that `payload` is omitted if the output of `payloadCreator` is `undefined`, but your list I think says something about omitting `payload` when `payloadCreator` itself is `undefined`.

Regardless, if the convention is to use `undefined` to omit an argument&mdash;and I think it should be&mdash;then it's probably a good idea to change the tests to follow that everywhere, even in cases where using `null` would have the same effect.

timche:
@jrasky the changes in this PR are already sufficient in your opinion?

timche:
@jrasky the changes in this PR are already sufficient in your opinion to fullfill your needs??

yangmillstheory:
>Regardless, if the convention is to use undefined to omit an argument—and I think it should be—then it's probably a good idea to change the tests to follow that everywhere, even in cases where using null would have the same effect.

I was thinking about saying this, but looks like it's already been said.

yangmillstheory:
***
@timche 

```js
// 4. Accept a second parameter as function for adding meta to the object
// 
// Returns:
// {
//   type: 'TYPE',
//   payload: 'foo',
//   meta: 'bar'
// }
const actionCreator = createAction(TYPE, (a) => a, (b) => b)
actionCreator('foo', 'bar')
```

I think this behavior is incorrect; `payloadCreator` and `metaCreator` have the same parameter signature, so `metaCreator `should be `(a, b) => b` in this case to produce the commented return value.

The examples spec looks 👌 .

To add to it, I don't like that we're coercing a non-function [here](https://github.com/acdlite/redux-actions/blob/e875b471aa2b1ed9262dc823ccdb7f9e876d5e22/src/createAction.js#L5). 

To add a bit more to your first example, I think that like elsewhere in the API, we should ask for a function or `undefined` for `payloadCreator` and `metaCreator`, and throw a `TypeError` otherwise. If given `undefined`, we should use `identity`.

See also https://github.com/acdlite/redux-actions/pull/97#discussion_r70165690.

However, this hardening doesn't belong in this PR; I can open up a separate one for it if you're on board.
***
@jrasky 
>Hmm, your list confuses me a little bit. This PR is such that payload is omitted if the output of payloadCreator is undefined, **but your list I think says something about omitting payload when payloadCreator itself is undefined.**

I don't see anything in the list about omitting payload when `payloadCreator` is `undefined`.

src/__tests__/createAction-test.js
```diff
@@ -95,7 +95,8 @@ describe('createAction()', () => {
 
       const explictNullAction = createAction(type)(null);
       expect(explictNullAction).to.deep.equal({
-        type
+        type,
+        payload: null
       });
 
       const baz = '1';
```
yangmillstheory: Can we get rid of this test? It doesn't have anything to do with the test name and is already being tested [here](https://github.com/acdlite/redux-actions/pull/128/files#diff-cd3494025bc48661a763dcb654f5a116R91).

yangmillstheory:
Regarding https://github.com/acdlite/redux-actions/pull/128#issuecomment-245724723, I don't see any tests where we're calling action creators with `null` and `undefined` where we're not intending to; in fact, with this change, I'm sure tests would fail in that case, as one test already had to be changed.

I **do** see calls to `createAction` with `null` `payloadCreator`, but I **don't** think this should change now. It should be changed (and will have to be changed, since tests will throw errors) in a separate PR, when only allow functions and `undefined` for `payloadCreator` and `metaCreator`. 

See https://github.com/acdlite/redux-actions/pull/128#issuecomment-245730284.

yangmillstheory:
@time Let's update the README to specify that the `payload` key will exist if and only if the `payloadCreator` returns a defined value.

After that, LGTM.

jrasky:
@yangmillstheory what about number five? I'm probably confused about how this relates to the other examples. Otherwise, something in the README would be perfect.

yangmillstheory:
@jrasky

Thanks for clarifying, I completely missed that; the fifth example is what you're referring to.

I think that example is specifying the wrong behavior; we shouldn't omit payload if given an `undefined` `payloadCreator`, but use the identity as the `payloadCreator` (and the `metaCreator`, since it's provided in that example as well).

This would require no new code changes, since we've always delegated to `identity`. So the corrected example would be

```js
// Returns:
// {
//   type: 'TYPE',
//   payload: 'bar',
//   meta: 'bar'
// }
const actionCreator = createAction(TYPE, undefined, (a) => a)
actionCreator('bar')
``` 

but is redundant with (and contradicts) the second example, which is correct. 

We should probably all be aligned on this point...

timche:
Thanks for the discussion guys.

I've fixed example 4 according to your comment @yangmillstheory:

```js
const TYPE = 'TYPE';
const errorObj = new TypeError('foo')

// 1. Throw an error when payloadCreator is not `undefined` or a function
// 
// Returns:
// new TypeError('payloadCreator must be undefined or a function')
const actionCreatorB = createAction(TYPE, 'foo')

// 2. Use identity function when payloadCreator is `undefined`
// 
// Returns:
// {
//   type: 'TYPE'
//   payload: 'foo'
// }
const actionCreator = createAction(TYPE)
actionCreator('foo')

// 3. Use payloadCreator when it is a function
// 
// Returns:
// {
//   type: 'TYPE',
//   payload: {
//     a: 'foo',
//     b: 'bar'
//   }
// }
const actionCreator = createAction(TYPE, (a, b) => ({ a, b }))
actionCreator('foo', 'bar')

// 4. Accept a second parameter as function for adding meta to the object
// 
// Returns:
// {
//   type: 'TYPE',
//   payload: 'foo',
//   meta: 'bar'
// }
const actionCreator = createAction(TYPE, (a) => a, (a, b) => b)
actionCreator('foo', 'bar')

// 5. Omit payload when payloadCreator is `undefined`, but a second parameter has been set for adding meta to the object
// 
// Returns:
// {
//   type: 'TYPE',
//   meta: 'bar'
// }
const actionCreator = createAction(TYPE, undefined, (a) => a)
actionCreator('bar')

// 6. Set `error` to `true` when payload is an Error object and bypass any payloadCreator
// 
// Returns:
// {
//   type: 'TYPE',
//   error: true,
//   payload: new TypeError('foo')
// }
const actionCreatorA = createAction(TYPE)
const actionCreatorB = createAction(TYPE, (a, b) => ({ a, b }))
actionCreatorA(errorObj)
actionCreatorB(errorObj)

// 7. Set error to true when payload is an Error object and bypass any payloadCreator, but not metaCreator
// 
// Returns:
// {
//   type: 'TYPE',
//   error: true,
//   payload: new TypeError('foo'),
//   meta: 'bar'
// }
const actionCreator = createAction(TYPE, (a) => a, (b) => b)
actionCreatorA(errorObj, 'bar')
```

Regarding example 5: It's not very clear to me then when to use `undefined`, when we are saying, [`undefined` should omit sth](https://github.com/acdlite/redux-actions/pull/115#issuecomment-237035075). How about we use `null` instead to make it more predictable that we don't want to define a `payloadCreator` function and just use the default one `a => a`? 

Also a more understandable example for 5 with `meta` would be to use the second parameter, otherwise I don't understand why `payload` and `meta` should have the same value :P

```js
// Returns:
// {
//   type: 'TYPE',
//   payload: 'bar',
//   meta: 'bar'
// }
const actionCreator = createAction(TYPE, null, (a, b) => b)
actionCreator('bar')
```js

**Comment by yangmillstheory** [comment]
***
1 sounds correct.

***
2

Why do we need to allow `null` as well when `undefined` already gives us the identity payload creator? Regardless, I can live with the extra complication, as long as `undefined` has the same behavior. 

Also, will we throw errors in the case that we receive a payload creator that is not `null`, `undefined`, or a `Function`? See https://github.com/acdlite/redux-actions/pull/129/.

**Comment by jrasky** [comment]
I would agree that using `undefined` is better when what's meant is "use a sane default", especially since actually omitting the argument results in `undefined`, but I agree with @yangmillstheory that that doesn't have to be the only behavior.

**Comment by timche** [comment]
> Also, will we throw errors in the case that we receive a payload creator that is not `null`, `undefined`, or a `Function`? See https://github.com/acdlite/redux-actions/pull/129.

Yes, please.

---

Alright, so lets agree the following for 2:

`undefined` as `payloadCreator` will use the identity function `a => a`.

---

Overall result now is:

1. Only `undefined` will omit `payload` when passed to the identity function.
2. `undefined` as `payloadCreator` will use the identity function `a => a`.

---

If we agree on these rules for `v1.0.0`, we are good too release it asap.

**Comment by yangmillstheory** [comment]
I'm OK with the proposed. ☝

**Comment by BerkeleyTrue** [comment]
Seems weird to use `undefined` as a parameter.

Isn't it convention to use `null` when something is omitted with purpose? Meaning intentionally left blank.

Where `undefined` is meant to be provided by the environment when a value is not found?

Also, there is the fact that undefined can be overwritten, causing this rule to exist http://eslint.org/docs/rules/no-undefined.

I've used this on all my projects for years now.

**Comment by jrasky** [comment]
@BerkeleyTrue take a look at the previous discussion on #115. The short of it is that treating `null` like `undefined` means you can't pass a payload of `null`, and that other projects (including redux itself) and the javascript standard treat the two differently.

It's no longer the case that you can change the value of `undefined`. According the the spec, `null` means "explicitly no value," whereas `undefined` means "no value (yet)."

**Comment by BerkeleyTrue** [comment]
@jrasky I see. Thanks for the clarification.

**Comment by yangmillstheory** [comment]
@timche is anything blocking this merge?

**Comment by timche** [comment]
@yangmillstheory Please approve once more, I've removed an obsolete test.

File: src/__tests__/createAction-test.js [hunk 2]
```diff
@@ -93,11 +93,6 @@ describe('createAction()', () => {
         type
       });
 
-      const explictNullAction = createAction(type)(null);
-      expect(explictNullAction).to.deep.equal({
-        type
-      });
-
```
yangmillstheory: Was this failing before?

Also, I'm wondering if there's a test somewhere that says the following:

```js
  const explictNullAction = createAction(type)(null);
  expect(explictNullAction).to.deep.equal({
  type,
  payload: null,
});

yangmillstheory:
@timche @jrasky @BerkeleyTrue 

I've been thinking more about Example 2, which eventually resulted in [this](https://github.com/acdlite/redux-actions/pull/128#issuecomment-246971375). I've been questioning it, wondering if it'll lead to more annoyance than usefulness and clarity.

However, I found some more material on the benefits of treating `null` and `undefined` differently (these are the first results that show up in a casual search)

- http://stackoverflow.com/a/2959845/2419669
- https://github.com/facebook/react/issues/2166#issuecomment-55040147
- https://github.com/graphql/graphql-js/issues/210
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters
- http://www.codereadability.com/javascript-default-parameters-with-or-operator/

So, I still think we're doing the right thing (there seems to be plenty of justification for it in the above material), even though I don't feel as strongly about it as before.

src/__tests__/createAction-test.js
timche: Yes, it was failing before, but I've removed it because of [your comment](https://github.com/acdlite/redux-actions/pull/128#r78083557) 😅

timche:
Thanks for the research @yangmillstheory, feeling more confident now that we are doing the change 😁

alexedev:
Also went in this problem a minute ago. 
I passed null to action explicitly and was expecting it to be in payload. For now decided to set value to null in reducer if it receives undefined payload, but hope this will be automated in future updates. Tank you for working on this issue! :-)

timche:
We are releasing `v1.0.0` very soon that allows `null` to be passed, thanks for your patience :)

timche:
[`v1.0.0` has been released!](https://github.com/acdlite/redux-actions/releases/tag/v1.0.0) :) Thanks for your patience!

PR #129, opened 2016-09-09
Check type of payload creator

yangmillstheory:
This takes care of the first example in the `createAction` spec outlined [here](https://github.com/acdlite/redux-actions/pull/128#issuecomment-244896176).

This contains breaking changes, so it might be worth including in https://github.com/acdlite/redux-actions/issues/122.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,9 @@ If you don’t use [npm](https://www.npmjs.com), you may grab the latest [UMD](h
 import { createAction } from 'redux-actions';
 ```
 
-Wraps an action creator so that its return value is the payload of a Flux Standard Action. If no payload creator is passed, or if it's not a function, the identity function is used.
+Wraps an action creator so that its return value is the payload of a Flux Standard Action. 
+
+`payloadCreator` must be a function or `undefined`. If `payloadCreator` is `undefined`, the identity function is used.
 
 Example:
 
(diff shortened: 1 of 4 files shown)


timche:
With the increase of checks and errors, I'd suggest to use [invariant](https://www.npmjs.com/package/invariant). It is already optimised for development and production with checking `process.env.NODE_ENV`.

yangmillstheory:
Using `invariant` per your suggestion now.

Can you explain the behavior when `process.env.NODE_ENV === 'production`? I'm not familiar with this library and couldn't quite figure it out via the README.

From my point of view the main benefit is reducing the boilerplate of throwing errors on certain conditions.

yangmillstheory:
For the `invariant` usage, I decided not to go with [this approach](https://github.com/ReactTraining/react-router/issues/545#issue-50437674), since the original error message is actually shorter than the "minified exception" message.

yangmillstheory:
@timche is it possible to change the base branch to `v1.0.0`?

timche:
Yes, I changed it.

BerkeleyTrue:
Any way we can allow `null` for payload creator? 

I've always used `null` when I want to omit an argument instead of undefined. I'm still hesitant to use `undefined`.

timche:
LGTM!

yangmillstheory:
@timche with this merge, [`v.1.0.0`](https://github.com/acdlite/redux-actions/projects/1) is good to go.

PR #130, opened 2016-09-09
Prefer "actionCreator" to "actionHandler" name

yangmillstheory:
By rights, `actionHandler` should be called `actionCreator`. 

This name is standard in the [Redux documentation](http://redux.js.org/docs/basics/Actions.html#action-creators) and community.

Also, "handler" in `actionHandler` collides with "handle" in `handleAction`, which is an avoidable point of confusion.


timche:
LGTM

PR #131, opened 2016-09-09
Enforce reducer type in handleAction

yangmillstheory:
Stop silently coercing bad reducer types; be clear on the API contract.

This is like #129, except for `handleAction` instead of `createAction`, so it has breaking changes.
---

Note that this doesn't make downstream errors happen earlier and more clearly (I checked this in tests in wrong reducer types), it just clarifies and enforces the API contract.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ expect(actionThree(3)).to.deep.equal({
 });
 ```
 
-### `handleAction(type, reducer | reducerMap, ?defaultState)`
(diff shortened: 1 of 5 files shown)


yangmillstheory:
@timche Should we change the base branch to `v1.0.0` since this has breaking changes?

timche:
Yep, changed here as well.

yangmillstheory:
Note to self: use `invariant` here as well.

PR #135, opened 2016-09-21
Remove lodash reduce and camelCase to save bytes

### LINKED ISSUE #134: redux-actions bundles lodash almost in its entirety
redux-actions.min.js is 32.4 kB, 82% of which is lodash:

screen shot 2016-09-20 at 4 21 26 pm

However, we only need the following things:

```
import camelCase from 'lodash/camelCase';
import identity from 'lodash/identity';
import includes from 'lodash/includes'
import isArray from 'lodash/isArray';
import isEmpty from 'lodash/isEmpty';
import isFunction from 'lodash/isFunction';
import isNil from 'lodash/isNil';
import isPlainObject from 'lodash/isPlainObject';
import isString from 'lodash/isString';
import isSymbol from 'lodash/isSymbol';
import reduce from 'lodash/reduce';
import toString from 'lodash/toString';
```

Looks like babel transforms these things just fine, only requiring individual files:

```
var _isString = require('lodash/isString');
```

I'm guessing maybe internally, lodash methods are just incredibly interconnected which causes such a hefty dependency weight.

lxe:
Before:

```
redux-actions.min.js  32.3 kB
```

After:

```
redux-actions.min.js   14 kB
```

Partially fixes #134

### DIFF
--- a/src/createActions.js
+++ b/src/createActions.js
@@ -1,5 +1,4 @@
 import identity from 'lodash/identity';
-import camelCase from 'lodash/camelCase';
 import isPlainObject from 'lodash/isPlainObject';
 import isArray from 'lodash/isArray';
 import reduce from 'lodash/reduce';
@@ -52,4 +51,12 @@ function fromIdentityActions(identityActions) {
       (actionsMap, actionType) => ({ ...actionsMap, [actionType]: identity })
     , {})
   );
+}
+
+function camelCase(word) {
+  return word.split(/[^a-zA-Z0-9]+/).map((part, index) => (
+    index === 0 ?
+      part.toLowerCase() :
+      part[0].toUpperCase() + part.slice(1).toLowerCase()
+  )).join('');
 }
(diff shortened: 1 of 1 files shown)


src/createActions.js
```diff
@@ -53,3 +53,11 @@ function fromIdentityActions(identityActions) {
...
   );
 }
+
+function camelCase(word) {
+  return word.split(/[^a-zA-Z0-9]+/).map((part, index) => (
+    index === 0 ?
+      part.toLowerCase() :
+      part[0].toUpperCase() + part.slice(1).toLowerCase()
+  )).join('');
+}
```
timche: Could you please write provide some tests as well? Just to make sure it works correctly and we don't break it in the future ;)

timche: ping @lxe

yangmillstheory:
I'm going to take care of this in #160, since there hasn't been any updates here. There's more incentive to take care of this given #159.

PR #136, opened 2016-09-22
set error to true when payload instanceof Error

xiaohanzhang:
Should set error to true when payloadCreator return error

### DIFF
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -133,12 +133,12 @@ describe('createAction()', () => {
 
     it('set error to true if payloadCreator return an Error object', () => {
       const errObj = new TypeError('this is an error');
-      const actionCreator = createAction(type, (...args) => errObj);
+      const actionCreator = createAction(type, () => errObj);
       const errAction = actionCreator('invalid arguments');
       expect(errAction).to.deep.equal({
         type,
         payload: errObj,
-        error: true,
+        error: true
       });
     });
   });
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -130,5 +130,16 @@ describe('createAction()', () => {
         meta: { foo: 'bar' }
       });
     });
+
+    it('set error to true if payloadCreator return an Error object', () => {
+      const errObj = new TypeError('this is an error');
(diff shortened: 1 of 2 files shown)


yangmillstheory:
@xiaohanzhang Thanks for the contribution!

- can you write some tests around this?
- can you describe some specific use cases (when I have an error, I usually pass the error to the action creator directly)?

yangmillstheory:
#43 is related.

@xiaohanzhang can you describe the use case?

@timche if the use case makes sense, I don't see a problem with merging this, as it seems to have been asked for and supported before.

xiaohanzhang:
@yangmillstheory Thanks for great work and quick reply, I added a test for it
use case:
```
const fooAction = createAction('FOO', (...args) => {
  return isValid(args) ? {foo: 'bar'} : new Error('arguments is not valid');
});
```
I feel it will make more sense to put arguments validator inside actions, otherwise we need check it in every components.
With current implementation, I need create another wrapper for real action:
```
const fooActionWrapper = (...args) => {
  return isValid(args) ? fooAction(...args) : fooAction(new Error());
}
```

yangmillstheory:
Thanks @xiaohanzhang. I think we should merge (pending @timche's feedback), but I would like to understand this a bit deeper, beyond avoiding writing wrappers. 

Can you give a couple of examples of an actual flow in your application where this comes in useful? What kind of validation are you doing in your payload creators?

xiaohanzhang:
We have some rest api action,  I want add more client-side validation before it send to server. 
It been used by many components and projects. So I hope can put the validator inside action.

yangmillstheory:
@timche please let us know if there are any blockers.

holybom:
Hi, this new change doesn't seem to affect payloadCreator that returns a Promise, there's no error field in the dispatched action. I'm using redux-promise as well.

`export const myAction = createAction('ACTION_TYPE', () => myAsyncFunc()); `

myAsyncFunc does some network requests and returns a Promise that can resolve into an Error. Can you suggest a way to make it work?

xiaohanzhang:
@holybom I usually just throw an error inside myAsyncFunc, redux-promise will catch that error:
https://github.com/acdlite/redux-promise/blob/master/src/index.js#L19

PR #139, opened 2016-09-27
handleAction should accept an empty action i.e., {}

CurtisHumphrey:
I write a lot of tests like the following which work until recent changes:
```javascript
describe('should have this initial state', () => {
  it('should start with no crumbs', () => {
    const next_state = reducer(undefined, {})
    expect(selectors.crumbs(next_state)).to.have.length(0)
  })
})

````

### DIFF
--- a/src/__tests__/handleAction-test.js
+++ b/src/__tests__/handleAction-test.js
@@ -19,6 +19,13 @@ describe('handleAction()', () => {
             counter: 7
           });
       });
+      it('returns default state if action is empty', () => {
+        const reducer = handleAction('NOTTYPE', () => null, { counter: 7 });
+        expect(reducer(undefined, {}))
+          .to.deep.equal({
+            counter: 7
+          });
+      });
 
       it('accepts single function as handler', () => {
         const reducer = handleAction(type, (state, action) => ({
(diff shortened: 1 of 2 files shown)


BerkeleyTrue:
An action, by definition, must have a type property. There is no such thing as an empty action. 

A way around this is just to create an action with `type: ''` and re-use it throughout your tests.

yangmillstheory:
Thanks @BerkeleyTrue. 

This library deals specifically with Flux Standard Actions. [The spec says that `type` is required](https://github.com/acdlite/flux-standard-action/#actions), though I'd say this is a broader definition in Flux/Redux.

CurtisHumphrey:
True I could create actions like {type:''}; however, redux itself handles actions of {} just fine so this library which is suppose to be a helper has artificial reduced redux abilities because it does not want to add about 20 chars: '''action.type == null || '''.  Also this library use to handle {} just fine until recent changes so to me this is adding in backwards compatibles since the version number did not indicate breaking changes. If you going to insist on this breaking change then please update the version numbers correctly.

JaKXz:
@CurtisHumphrey since `redux-actions` is v0.x, any minor version bump should be considered breaking according to the semver spec. 

That being said, I do think it's a valid fix to handle the empty object action since that matches Redux.

JaKXz:
However, the spec for FSAs mandate a type property and that's fair enough for this library to enforce. Thanks again @yangmillstheory! :)

yangmillstheory:
Just to clarify the intent of this library:

image

It should be used for apps that are compliant to the FSA spec. As [mentioned above](https://github.com/acdlite/redux-actions/pull/139#issuecomment-250378679), actions with no `type` are not FSA.

If you'd like to retain most of the functionality of this library, but not be FSA-compliant, you're free to fork.

Issue #126, opened 2016-09-01
Can we safely drop the Symbols support?


timche:
With v1.0.0 we would also like to drop the Symbols support as they are not recommended and ["generally a bad idea"](https://github.com/acdlite/redux-actions/issues/20#issuecomment-134340125).

Before we are dropping it completely, is anyone using them extensively?

yangmillstheory:
FYI [this is the original thread](https://github.com/acdlite/redux-actions/issues/20) where we decided to adopt Symbol support.

theghostbel:
You can put somewhere a "patch" so anyone can adopt "Symbal" supoprt if they need it. I personally don't know any project who uses Symbols.

JaKXz:
I'm curious to see if there are any use cases for Symbols before just dropping them. From the Redux docs:

> Note that it is okay to use Symbols, Promises, or other non-serializable values in an action if the action is intended for use by middleware. Actions only need to be serializable by the time they actually reach the store and are passed to the reducers.

but it might make _more_ sense for FSA to "standardize" on strings only for the [advantages mentioned](http://redux.js.org/docs/faq/Actions.html#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants).

timche:
If we are unsure about it, I think it's okay to keep it. They don't do anything harmful.

JaKXz:
@timche I agree - I think it makes a better candidate for the next major version to be dropped. We can mark it as deprecated for now and see if there's pushback.

PR #141, opened 2016-10-06
fix(handleAction): Add descriptive error for missing or invalid actions

### LINKED ISSUE #145: Fail early for non-FSA actions
Since this library is for Redux users compliant to [FSA](https://github.com/acdlite/flux-standard-action/), we should fail explicitly when `handleAction` is called with a non-FSA.

See #141, https://github.com/acdlite/redux-actions/pull/139#issuecomment-250376196, and https://github.com/acdlite/redux-actions/issues/140.

JaKXz:
@yangmillstheory this might be redundant or overkill so feel free to close - but it might be helpful to folks like me to see a more descriptive error when we forget the action type (with the helpful suggestion to use `creatAction` or `createAction(s)`)

~~I also took the liberty of adding an `.editorconfig` and ignoring the JetBrains `.idea` folder. And removing Node 5 from testing since it's past it's useful life. If any of these bother you I can quickly rebase this and take those changes out.~~

Resolves #145.

### DIFF
--- a/src/__tests__/handleAction-test.js
+++ b/src/__tests__/handleAction-test.js
@@ -204,6 +204,7 @@ describe('handleAction()', () => {
         (state) => state
       );
       expect(() => reducer(undefined))
+      // eslint-disable-next-line max-len
         .to.throw(/The FSA spec mandates an action object with a type. Try using the createAction\(s\) method./);
     });
 
@@ -214,6 +215,7 @@ describe('handleAction()', () => {
         (state) => state
       );
       expect(() => reducer(undefined, {}))
+      // eslint-disable-next-line max-len
         .to.throw(/The FSA spec mandates an action object with a type. Try using the createAction\(s\) method./);
     });
 
@@ -224,6 +226,7 @@ describe('handleAction()', () => {
         (state) => state
       );
       expect(() => reducer(undefined, { type: false }))
+      // eslint-disable-next-line max-len
         .to.throw(/The FSA spec mandates an action object with a type. Try using the createAction\(s\) method./);
     });
   });
--- a/src/__tests__/handleAction-test.js
+++ b/src/__tests__/handleAction-test.js
@@ -200,23 +200,32 @@ describe('handleAction()', () => {
   describe('with invalid actions', () => {
     it('should throw a descriptive error when the action object is missing', () => {
       const reducer = handleAction(createAction('ACTION_1'), identity);
-      expect(() => reducer(undefined))
-      // eslint-disable-next-line max-len
-        .to.throw(/The FSA spec mandates an action object with a type. Try using the createAction\(s\) method./);
+      expect(
+        () => reducer(undefined)
+      ).to.throw(
+        Error,
+        'The FSA spec mandates an action object with a type. Try using the createAction(s) method.'
+      );
     });
 
     it('should throw a descriptive error when the action type is missing', () => {
       const reducer = handleAction(createAction('ACTION_1'), identity);
-      expect(() => reducer(undefined, {}))
-      // eslint-disable-next-line max-len
-        .to.throw(/The FSA spec mandates an action object with a type. Try using the createAction\(s\) method./);
+      expect(
+        () => reducer(undefined, {})
+      ).to.throw(
+        Error,
+        'The FSA spec mandates an action object with a type. Try using the createAction(s) method.'
+      );
     });
 
     it('should throw a descriptive error when the action type is not a string or symbol', () => {
       const reducer = handleAction(createAction('ACTION_1'), identity);
-      expect(() => reducer(undefined, { type: false }))
-      // eslint-disable-next-line max-len
-        .to.throw(/The FSA spec mandates an action object with a type. Try using the createAction\(s\) method./);
+      expect(
+        () => reducer(undefined, { type: false })
+      ).to.throw(
+        Error,
+        'The FSA spec mandates an action object with a type. Try using the createAction(s) method.'
+      );
     });
   });
 });
--- a/src/__tests__/handleAction-test.js
+++ b/src/__tests__/handleAction-test.js
@@ -195,4 +195,36 @@ describe('handleAction()', () => {
         .to.deep.equal({ number: 3 });
     });
   });
+
+  describe('with invalid actions', () => {
+    it('should throw a descriptive error when the action object is missing', () => {
+      const action1 = createAction('ACTION_1');
+      const reducer = handleAction(
+        action1,
+        (state) => state
+      );
+      expect(() => reducer(undefined))
+        .to.throw(/The FSA spec mandates an action object with a type. Try using the createAction\(s\) method./);
+    });
+
+    it('should throw a descriptive error when the action type is missing', () => {
+      const action1 = createAction('ACTION_1');
+      const reducer = handleAction(
+        action1,
+        (state) => state
+      );
+      expect(() => reducer(undefined, {}))
+        .to.throw(/The FSA spec mandates an action object with a type. Try using the createAction\(s\) method./);
+    });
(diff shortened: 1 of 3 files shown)


src/handleAction.js
```diff
@@ -12,10 +14,18 @@ export default function handleAction(actionType, reducers, defaultState) {
...
+
     if (!includes(actionTypes, action.type.toString())) {
       return state;
     }
 
     return (action.error === true ? throwReducer : nextReducer)(state, action);
   };
 }
+
+function isValidFSA(action) {
```
yangmillstheory: Can we use https://github.com/acdlite/flux-standard-action/#isfsaaction?

src/handleAction.js
```diff
@@ -12,10 +14,18 @@ export default function handleAction(actionType, reducers, defaultState) {
     : [reducers.next, reducers.throw].map(reducer => (isNil(reducer) ? identity : reducer));
 
   return (state = defaultState, action) => {
+    if (!isValidFSA(action)) {
```
yangmillstheory: Can we standardize on [`invariant`](https://www.npmjs.com/package/invariant)?

.editorconfig
```diff
@@ -0,0 +1,16 @@
+root = true
```
yangmillstheory: I'm not familiar with `.editorconfig`. Can you quickly teach me what value this adds?

JaKXz: Sure! [EditorConfig](http://editorconfig.org/) is really good for keeping line endings, indentation, trailing spaces and so on and so forth in sync between different IDEs. It's just a quick headache reducer as long as contributors have it installed in their editors (some of which have it bundled out of the box).

src/handleAction.js
JaKXz: Ah - yes. Will change.

yangmillstheory: Sounds good. Though, we may be dropping Symbol support soon https://github.com/acdlite/redux-actions/issues/126.

.editorconfig
JaKXz: It doesn't do linting, it just works inside the editor make those things happen (eg. tabs = 2 spaces, or all files get a newline at the end).

JaKXz:
Upstream PR for enforcing the `action.type`'s type: https://github.com/acdlite/flux-standard-action/pull/36

yangmillstheory:
Could we make a separate PR and issue for the unrelated changes (adding `.editorconfig` and removing node@v5.x.x from the build? 

That would help clarify the scope of this PR and make it more targeted.

JaKXz:
@yangmillstheory made separate PRs :) #142 and #143.

.gitignore
```diff
@@ -3,3 +3,4 @@ coverage
 *.log
 lib
 dist
+.idea
```
timche: Can you create a single commit or pr for this as well pls?

JaKXz: @timche #144 :)

src/handleAction.js
JaKXz: done :)

src/__tests__/handleAction-test.js
```diff
@@ -195,4 +195,39 @@ describe('handleAction()', () => {
         .to.deep.equal({ number: 3 });
     });
   });
+
+  describe('with invalid actions', () => {
+    it('should throw a descriptive error when the action object is missing', () => {
+      const action1 = createAction('ACTION_1');
+      const reducer = handleAction(
+        action1,
+        (state) => state
```
yangmillstheory: For this and the tests below:

- can we use the identity function from lodash?
- can we inline the call to `createAction` (`action1` is not used elsewhere)?

Also, unless there's a good reason to disable`max-len`, let's stick to it. Since we're testing an exact match, can we use the string directly?

yangmillstheory:
The build is failing. I'd like to get this in, so let's make sure that passes as soon as we can.

JaKXz:
@yangmillstheory I've made https://github.com/acdlite/flux-standard-action/pull/36 pass CI, but it needs to be merged and released to unblock CI here.

yangmillstheory:
We're planning on dropping Symbol support soon #126 (please see the linked discussions there). I'm not sure it's worth waiting on upstream support for Symbols.

JaKXz:
@yangmillstheory good point, but upstream only checks if `action.type` is truthy, not a string. That's why the last test is failing.

It will be easy enough to remove support for Symbols in the next major (breaking) version. :)

yangmillstheory:
I'm not sure that the last test (or the upstream PR) is a requirement for this merge. While I agree it's _technically_ correct to enforce that the action type is a string (per the FSA spec IIRC), anyone who would explicitly pass `false` for an action type might reasonably be referred to library documentation.

What other tests are failing beyond that one?

JaKXz:
@yangmillstheory that's the only test that's failing, but I would say that the explicit type-check is blocking since it's not spec-compliant. I think I see what you're saying about being able to merge without that test passing since it's not necessarily a "valid use case" (i.e. nobody will actually do it) but I'd personally rather be explicit first.

Up to you all as maintainers though.

JaKXz:
@yangmillstheory @timche I've published my fork of `flux-standard-action` (the irony is strong with this one) since @acdlite has not gotten back to me.

This should be ready to merge now. :)

JaKXz:
@yangmillstheory @timche updated :) _really_ ready to merge now.

src/__tests__/handleAction-test.js
```diff
@@ -195,4 +196,27 @@ describe('handleAction()', () => {
         .to.deep.equal({ number: 3 });
     });
   });
+
+  describe('with invalid actions', () => {
+    it('should throw a descriptive error when the action object is missing', () => {
+      const reducer = handleAction(createAction('ACTION_1'), identity);
+      expect(() => reducer(undefined))
+      // eslint-disable-next-line max-len
+        .to.throw(/The FSA spec mandates an action object with a type. Try using the createAction\(s\) method./);
```
yangmillstheory: I'd like to push back again on the `eslint-disable-next-line max-len` and the use of a regular expression here.

I'd like to use these overrides only where they must be used. In a 20K line production application, I've seen this happen only in two places:

* directly consuming snake-cased fields from the API on the client (`eslint-disable-next-line camelcase`)
* early returns from iterators in `lodash` (`eslint-disable-next-line consistent-return`)

It wouldn't be a rule if we were permissive about bypassing it; the value of the rule comes from a widespread adherence.

***

I also think we should strive for consistency with respect to [similar tests in the repo](https://github.com/acdlite/redux-actions/blob/master/src/__tests__/createActions-test.js#L17). 

To that end, let's test the error class thrown, so callers know what to catch, and the exact string (I don't think the flexibility afforded by using a regular expression object buys us much).

JaKXz: Done. Thanks for the feedback!

`invariant` only throws a regular `Error` and I didn't want to introduce matchers just to check for the error name to be "Invariant Violation". I didn't see it being much value, but the format should be consistent now.

yangmillstheory:
I'm "dismissing" @timche's review since all asks have been addressed. I'm also changing the merge target branch from `master` to `v1.0.0`.

In the future, let's strive to create more targeted PR's with smaller diffs. That will ensure that reviews are more focused and less noisy.

Thanks for the contribution!

yangmillstheory:
@JaKXz can you fetch the upstream `v1.0.0` branch, merge those changes into your branch, and update the PR?

JaKXz:
@yangmillstheory done. Anything I can do to help #127 get merged?

yangmillstheory:
Thanks. I'll take care of #127 next. 

Now that you're a collaborator on FSA, do you have any opinions on #126?

PR #142, opened 2016-10-06
chore: add editorconfig

JaKXz:


.editorconfig
```diff
@@ -0,0 +1,16 @@
+root = true
+
+[*]
+
```
timche: Please remove this whitespace as well for consistent style like L15-16.

PR #143, opened 2016-10-06
chore(travis): Test against latest Node LTS versions

JaKXz:
Since Node 5 is past its useful life, speed up the build process by just
specifying LTS versions 4 and 6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: node_js
 node_js:
   - "6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: node_js
 node_js:
   - "6"
   - "4"
-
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,8 +1,9 @@
+sudo: false
```
timche: `sudo` is by default `false` for repos since 2015 on Travis, also see log.

.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: node_js
 node_js:
   - "6"
-  - "5"
   - "4"
+
```
timche: Any reasons for this whitespace? If not, I'd suggest to remove it as well to stay on the context.

timche: No problem, thanks ;)

PR #147, opened 2016-10-14
allow metaCreator for an object

bshy522:
A shortened form of metaCreator.

Thanks.


src/__tests__/createAction-test.js
```diff
@@ -48,6 +48,20 @@ describe('createAction()', () => {
       expect(isFSA(action)).to.be.true;
     });
 
+    it('allow metaCreator for an object', () => {
+      const actionCreator = createAction(type, null, { cid: 5 });
+      const foobar = { foo: 'bar', cid: 5 };
```
yangmillstheory: Please name this `payload` instead of `foobar`.

bshy522: Thank you for your proposal. I've fixed it.

yangmillstheory:
@timche can you take a look?

@bshy522 I think this looks good. I'm curious about use case though, as it's interesting to see how people use the library in the wild. What problem does this solve for you?

JaKXz:
@yangmillstheory this would be useful since I've found myself writing a lot of functions that look like `() => ({ my: "meta object"})` so it feels a bit redundant.

yangmillstheory:
@JaKXz From the README:

>It receives the same arguments as the payload creator, but its result becomes the meta field of the resulting action.

yangmillstheory:
Paging @timche.

I'd still be interested to know of an application-specific example of how this is used. Not trying to question the use case, just curiosity.

JaKXz:
@yangmillstheory ah - must've missed that. Thanks!

yangmillstheory:
Thanks for the contribution and very sorry for the lack of movement. I would like to get this in over the weekend. I'll let it sit and see if anyone has any thoughts, and then merge into `master`.

README.md
```diff
@@ -83,7 +83,7 @@ Use the identity form to create one-off actions:
 createAction('ADD_TODO')('Use Redux');
 ```
 
-`metaCreator` is an optional function that creates metadata for the payload. It receives the same arguments as the payload creator, but its result becomes the meta field of the resulting action. If `metaCreator` is undefined or not a function, the meta field is omitted.
+`metaCreator` is an optional object or function which creates metadata for the payload. It receives the same arguments as the payload creator, but its result becomes the meta field of the resulting action. If `metaCreator` is undefined or not a function or an object, the meta field is omitted. When `metaCreator` is an object, it will be returned directly as metadata.
```
timche: We could rephrase this paragraph to something like this, so we don't switch between object and function contexts few times:

"`metaCreator` is an optional argument that can either be an object or a function. An object will be directly used as `meta` value in the payload, whereas a function receives the same arguments as the payload creator and its return value will be used as `meta` value. If `metaCreator` is not defined, the `meta` property will be omitted."

bshy522: I'm coming a bit late.
My English is so bad, thank you very much.

JaKXz:
Looks like this is ready to go now? :)

timche:
@bshy522 sorry for not having this pushed for such a long time. Still available to update this PR? Would be happy to merge this.

alexander-heimbuch:
Closing this since the changes already have been integrated.

PR #161, opened 2016-11-18
Support Additional arguments for handleAction(s)

### LINKED ISSUE #148: Additional arguments for handleAction{,s} reducers
A feature request here.

I have a reducer which have cross-cutting concerns, basically an action logger which puts log lines in the store based on different kinds of actions. This logger needs to know about a broader part of the state than the list of log statements its reducing. By sequencing the reducers in the root reducer, the log reducer can get for example the previous and next version of the state and log accordingly.

Having a root reducer similar to the following (using Immutable containers):

``` javascript
export default (state = new Immutable.Map(), action) => (
  state
    .update('someState', s => someState(s, action))
    .update(next => next.update('log', l => log(l, action, state, next)))
);
```

I'd optimally want to do the following:

``` javascript
const logInitialState = new Immutable.List();
const log = handleActions({
  // Direct logging action
  [logAppend]: (state, action) => state.push(Immutable.fromJS(action.payload)),

  // React to some other non-logging related action, notice the extra args
  [someAction]: (state, action, prev, next) => {
    const value = next.get('someState');
    if (value === prev.get('someState')) return state;
    return state.push(Immutable.fromJS({
      type: 'someAction',
      message: 'someAction happened',
    }));
  },
}, logInitialState);
```

But since handleActions does not relay the extra arguments, I'm forced to generate the reducers each time and close around the extra arguments like so:

``` javascript
const logInitialState = new Immutable.List();
const log = (state, action, prev, next) => handleActions({
  // Direct logging action
  [logAppend]: () => state.push(Immutable.fromJS(action.payload)),

  // React to some other non-logging related action
  [someAction]: () => {
    const value = next.get('someState');
    if (value === prev.get('someState')) return state;
    return state.push(Immutable.fromJS({
      type: 'someAction',
      message: 'someAction happened',
    }));
  },
}, logInitialState)(state, action);
```

This is both extra verbose and inefficient. It would be great if both `handleAction` and `handleActions` returns reducers that fetch a rest spread and passes it on. I guess `reduce-reducers` needs to know about this as well.

Something like the following would probably do:

``` javascript
export default function handleActions(handlers, defaultState) {
  // snip
  return (state = defaultState, action, ...args) => reducer(state, action, ...args);
}
```

xiaohanzhang:
* resolve #148

### DIFF
--- a/src/__tests__/handleAction-test.js
+++ b/src/__tests__/handleAction-test.js
@@ -65,6 +65,22 @@ describe('handleAction()', () => {
             counter: 10
           });
       });
+
+      it('should should support additional arguments', () => {
+        const { increment } = createActions('INCREMENT');
+
+        const reducer = handleAction(
+          increment, (state, { payload }, extraArg) => ({
+            counter: state.counter + payload + extraArg
+          }), 
+          { counter: 3 });
+
+        const counterBase = 10;
+        expect(reducer(undefined, increment(7), counterBase))
+          .to.deep.equal({
+            counter: 10 + counterBase
+          });
+      });
     });
   });
 
--- a/src/__tests__/handleActions-test.js
+++ b/src/__tests__/handleActions-test.js
@@ -147,4 +147,21 @@ describe('handleActions', () => {
         counter: 7
       });
   });
+
+  it('accepts additional arguments', () => {
+    // meaningless base and fix just for testing additional arguments
+    const incrementAction = createAction('INCREMENT');
+    const reducer = handleActions({
+      [incrementAction]: (
+        { counter }, { payload: amount }, extraArg1, extraArg2
+      ) => ({ counter: counter + amount + extraArg1 + extraArg2 })
+    });
+
+    const counterBase = 5;
+    const counterFix = 1;
+    expect(reducer({ counter: 3 }, incrementAction(7), counterBase, counterFix))
+      .to.deep.equal({
+        counter: 10 + counterBase + counterFix
+      });
+  });
 });
--- a/src/handleActions.js
+++ b/src/handleActions.js
@@ -1,10 +1,11 @@
 import handleAction from './handleAction';
 import ownKeys from './ownKeys';
-import reduceReducers from 'reduce-reducers';
 
 export default function handleActions(handlers, defaultState) {
   const reducers = ownKeys(handlers).map(type => handleAction(type, handlers[type]));
-  const reducer = reduceReducers(...reducers);
+  const reducer = (state, action, ...args) => reducers.reduce(
+    (p, r) => r(p, action, ...args), state);
 
-  return (state = defaultState, action) => reducer(state, action);
+  return (state = defaultState, action, ...args) => reducer(state, action, ...args);
 }
+
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -11,11 +11,13 @@ export default function handleAction(actionType, reducers, defaultState) {
     ? [reducers, reducers]
     : [reducers.next, reducers.throw].map(reducer => (isNil(reducer) ? identity : reducer));
 
-  return (state = defaultState, action) => {
+  return (state = defaultState, action, ...args) => {
     if (!includes(actionTypes, action.type.toString())) {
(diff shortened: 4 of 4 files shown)


src/__tests__/handleActions-test.js
```diff
@@ -147,4 +147,21 @@ describe('handleActions', () => {
         counter: 7
       });
   });
+
+  it('accepts additional arguments', () => {
+    // meaningless base and fix just for testing additional arguments
```
yangmillstheory: Is this comment necessary?

src/__tests__/handleAction-test.js
```diff
@@ -65,6 +65,22 @@ describe('handleAction()', () => {
...
       });
+
+      it('should should support additional arguments', () => {
+        const { increment } = createActions('INCREMENT');
+
+        const reducer = handleAction(
+          increment, (state, { payload }, extraArg) => ({
+            counter: state.counter + payload + extraArg
+          }),
+          { counter: 3 });
```
yangmillstheory: Can we put the closing parenthesis on a newline, and use the same number of arguments in `handleActions-test.js`?

src/__tests__/handleActions-test.js
```diff
@@ -147,4 +147,21 @@ describe('handleActions', () => {
...
+    const incrementAction = createAction('INCREMENT');
+    const reducer = handleActions({
+      [incrementAction]: (
+        { counter }, { payload: amount }, extraArg1, extraArg2
+      ) => ({ counter: counter + amount + extraArg1 + extraArg2 })
+    });
+
+    const counterBase = 5;
+    const counterFix = 1;
```
yangmillstheory: Can we use `extraArg1` and `extraArg2` instead, to mirror the parameter names you used above?

src/__tests__/handleActions-test.js
yangmillstheory: Is this comment necessary?

src/handleActions.js
```diff
@@ -1,10 +1,11 @@
 import handleAction from './handleAction';
 import ownKeys from './ownKeys';
-import reduceReducers from 'reduce-reducers';
 
 export default function handleActions(handlers, defaultState) {
   const reducers = ownKeys(handlers).map(type => handleAction(type, handlers[type]));
-  const reducer = reduceReducers(...reducers);
+  const reducer = (state, action, ...args) => reducers.reduce(
+    (p, r) => r(p, action, ...args), state);
```
yangmillstheory: I don't like these one-letter variable names.

Can we please use: 

```js
reducers.reduce((previousState, reducer) => reducer(previousState, action, ...args), state)
```

Also, at this point, we should remove `reducer-reducers` from package.json, since we're no longer using it. 

And, could you submit a pull request [here](https://github.com/acdlite/reduce-reducers) for passing the extra arguments?

src/__tests__/handleAction-test.js
```diff
@@ -65,6 +65,22 @@ describe('handleAction()', () => {
...
+
+        const counterBase = 10;
```
yangmillstheory: Let's call this `extraArg`; `counterBase` doesn't really signify anything.

yangmillstheory:
Can we update the README to indicate that extra arguments are passed?

Thanks for the contribution. I'd like to wait over the weekend to see if @timche has any thoughts, and then merge.

yangmillstheory:
Sorry, we just merged `v1.0.0`. Can you resolve conflicts?

yangmillstheory:
@timche will take a look tomorrow.

timche:
Before we add this feature, I'd like to discuss this a bit deeper: https://github.com/acdlite/redux-actions/issues/148

lsunsi:
I think this change is good. I realize it violates reducer signature if the users use the additional arguments, but from what I can gather, this is a great way of passing global state as a read only to a slice reducer that can write to it's own slice.

The pattern seems fine and the change seems harmless, since it would change literally anything, it'll just be less opinionated on how you use yours reducers which are, as we know, `just functions`. 

+1 for the change

Issue #159, opened 2016-11-12
Prefixes/Namespacing proposed syntax


Strandedpirate:
For namespacing, which I would like to see, how about supporting this kind of syntax? This actually reduces boilerplate in the destructing assignment for all 2'nd level action creators and prevents having to keep keys and their destructed names in sync.

```js
const {
  rootLevelAction,
   // no keys below products or orders must be kept in sync here. yeah!!!
  products,
  orders,
} = createActions({
	rootLevelAction: () => ({}),
	products: { // this key used for the namespace prefix
	  list: (query) => ({ query }),
	  addToCart: (quantity, productId) => ({ quantity, productId }),
	},
	orders: { // this key used for the namespace prefix
	  placeOrder: (cartId) => ({ cartId }),
	  getOrder: (orderId) => ({ orderId }),
	}
});
```
	
Example usage:
```js
dispatch(rootLevelAction());
dispatch(products.list('Star Wars'));
dispatch(products.addToCart(1, 45243));
dispatch(orders.placeOrder(7348));
dispatch(orders.getOrder(45));
```

Destructures to the following:
```js
rootLevelAction: function() {
	return {
		type: 'rootLevelAction',
		payload: { }
	};
},
products:{
	type: 'products',
	list: function(query) {
		return {
			type: 'products/list', // boom, namespaced to reduce collisions
			payload: { query }
		};
	},
	addToCart: function(quantity, productId) {
		return {
			type: 'products/addToCart',
			payload: { quantity, productId }
		};
	}
},
orders:{
	type: 'orders',
	placeOrder: function(cartId) {
		return {
			type: 'orders/placeOrder',
			payload: { cartId }
		};
	},
	getOrder: function(orderId) {
		return {
			type: 'orders/getOrder',
			payload: { orderId }
		};
	}
}
```

Supporting an unlimited level of literal nesting would be awesome and allow for richly namespaced action creators that reduce the chance of collisions.

yangmillstheory:
I would prefer that this be kept out of the library; but that doesn't mean we can't come up with a reasonable solution.

If we can re-implement a smarter camel-casing function (see #135), we can actually push the rest of the solution out to userland.

```js
const {createActions} = require('redux-actions')

// generic, should be pulled from a utility module
const flatten = (
  target,
  delimiter,
  flattened = {},
  path = ''
) => {
  function getNextPath(key) {
    return path ? `${path}${delimiter}${key}` : key
  }
  for (const key in target) {
[...]

PR #166, opened 2016-11-19
Remove lodash.reduce, lodash.camelCase

yangmillstheory:
Going from

```
redux-actions git:(master) npm run build:umd:min

> redux-actions@1.2.0 build:umd:min /Users/VictorAlvarez/code/private/redux-actions
> cross-env NODE_ENV=production webpack

Hash: 423ecf4d346ccf00b655
Version: webpack 1.13.1
Time: 2527ms
               Asset     Size  Chunks             Chunk Names
redux-actions.min.js  33.1 kB       0  [emitted]  main
    + 157 hidden modules
redux-actions git:(master)
```

to

```
redux-actions git:(camel-case) npm run build:umd:min

> redux-actions@1.2.0 build:umd:min /Users/VictorAlvarez/code/private/redux-actions
> cross-env NODE_ENV=production webpack

Hash: 8168e6ff577773b7fc11
Version: webpack 1.13.1
Time: 1856ms
               Asset     Size  Chunks             Chunk Names
redux-actions.min.js  17.3 kB       0  [emitted]  main
    + 67 hidden modules

```

Also, I think it's useful to allow `/` when camel-casing (and not remove it). This would give users the ability to write namespaced actions (via [helper code](https://github.com/yangmillstheory/redux-scoped-actions), e.g.) as in https://github.com/acdlite/redux-actions/issues/159.


src/camelCase.js
```diff
@@ -0,0 +1,14 @@
...
+
+function camelCase(string) {
+  return string.match(wordPattern).reduce((camelCased, word, index) =>
+    camelCased + (index === 0
+      ? word.toLowerCase()
+      : word.charAt(0).toUpperCase() + word.substring(1).toLowerCase())
+  , '');
+}
+
+export default actionType => actionType.split(namespacer).map(camelCase).join(namespacer);
```
timche: Let's use the camelCase function from sindresorhus: https://github.com/sindresorhus/camelcase

We could also put this file into a `utils` folders.

yangmillstheory: That doesn't look like it honors delimiters like `/`. From the PR description

>Also, I think it's useful to allow / when camel-casing (and not remove it). This would give users the ability to write namespaced actions (via helper code, e.g.) as in #159.

Thoughts?

timche: Fine for me. We should only look if we really covered every possible case.

ruiaraujo:
What is the progress on this? Looks good.

yangmillstheory:
Sorry about the delay! I agree this is mostly done, I'll update this and ship tonight.

yangmillstheory:
https://github.com/acdlite/redux-actions/releases/tag/1.2.1 @ruiaraujo thanks for moving this along

Issue #167, opened 2016-11-21
v1.0.0 change too much


geminiyellow:
i use something like this,

```
handleActions({}, {})
```

but when i update to v1.0.0, it die.

and i use `redux-api-middleware`, create actions like 

```
{
  [CALL_API]: {
    type: [xxx]
  }
}
```

but when i update to v1.0.0, it die.

@acdlite i think you can find a more smoothly way to update, not like this.

Lexicality:
What errors are thrown?

timche:
Thanks for reporting.

I assume you are getting [this error](https://github.com/acdlite/redux-actions/blob/master/src/handleAction.js#L27-L30)?

timche:
#168 has been opened to resolve this issue.

timche:
We have published [v1.0.1](https://github.com/acdlite/redux-actions/releases/tag/v1.0.1).

geminiyellow:
@timche thanks , and sorry i post 2 issues in 1

`export const shared = handleActions({}, {})`

```
Uncaught Error: Reducer "shared" returned undefined during initialization. If the state passed to the reducer is undefined, you must explicitly return the initial state. The initial state may not be undefined.
```

timche:
What exactly trying to achieve with `export const shared = handleActions({}, {})`? The function signature is [`handleAction(type, reducer | reducerMap = Identity, defaultState)`](https://github.com/acdlite/redux-actions#handleactiontype-reducer--reducermap--identity-defaultstate). The default state is missing is [required in Redux](http://redux.js.org/docs/basics/Reducers.html).

geminiyellow:
i just left the placeholder in the empty `reducer` file.
before v1.0.0, 

i can use
```
const shared = handleActions({}, {})

const reducers = {  shared, ui, ... };
const reducer = combineReducers(reducers);
```

create a empty state in store.
but now i cannot do this.

:trollface:

timche:
As mentioned before, a default state is required in Redux, regardless of redux-actions. Also see: https://github.com/acdlite/redux-actions/issues/18

yangmillstheory:
> Uncaught Error: Reducer "shared" returned undefined during initialization. If the state passed to the reducer is undefined, you must explicitly return the initial state. The initial state may not be undefined.


[...]

PR #168, opened 2016-11-21
Remove FSA check in handleAction(s)

### LINKED ISSUE #164: Non-FSA actions cause handleAction reudcers to emit confusing error messages

timche:
With #141 we have made other libraries incompatible with redux-actions and it resulted to an unexpected behavior (#164, #167). If an action is not a FSA, handleAction should not handle it and just pass it to the next reducer. This fix will remove the FSA check to support Non-FSA.

Closes #164, #167

### DIFF
--- a/src/__tests__/handleAction-test.js
+++ b/src/__tests__/handleAction-test.js
@@ -1,5 +1,4 @@
 import { expect } from 'chai';
-import identity from 'lodash/identity';
 import { handleAction, createAction, createActions, combineActions } from '../';
 
 describe('handleAction()', () => {
@@ -95,6 +94,20 @@ describe('handleAction()', () => {
           .to.deep.equal({
             counter: 7
           });
+      });
+
+      it('should not throw and return state when action is non-FSA', () => {
+        const reducer = handleAction(type, (state) => state, defaultState);
+        const action = {
+          foo: {
+            bar: 'baz'
+          }
+        };
+
+        expect(reducer(undefined, action)).not.to.throw;
+        expect(reducer(undefined, action)).to.deep.equal({
+          counter: 0
+        });
       });
     });
   });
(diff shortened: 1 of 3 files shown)


src/handleAction.js
```diff
@@ -24,12 +23,7 @@ export default function handleAction(actionType, reducer = identity, defaultStat
     : [reducer.next, reducer.throw].map(aReducer => (isNil(aReducer) ? identity : aReducer));
 
   return (state = defaultState, action) => {
-    invariant(
-      isFSA(action),
-      'The FSA spec mandates an action object with a type. Try using the createAction(s) method.'
-    );
-
-    if (!includes(actionTypes, action.type.toString())) {
```
yangmillstheory:
```js
const {type: actionType} = action
```

or 
```js
const {type} = action
```

timche:
We didn't made any changes when we added the check. The current README makes it pretty clear that `handleAction` only handles FSA.

src/handleAction.js
timche: Changed it.

src/__tests__/handleAction-test.js
```diff
@@ -96,6 +95,18 @@ describe('handleAction()', () => {
...
+      it('should work with non-FSA', () => {
+        const reducer = handleAction(type, (state) => state, defaultState);
+
+        expect(reducer(undefined, {
+          foo: {
+            bar: 'baz'
+          }
+        })).to.deep.equal({
+          counter: 0
+        });
```
yangmillstheory: Is there a chai assertion that we expect this code block not to throw? I think that's the point, and not that the state is reduced to a certain value.

yangmillstheory: http://chaijs.com/api/bdd/#method_throw

`.not.to.throw`

timche: This reducer should not throw **AND** just return the state. Will change that.

timche:
Please review again @yangmillstheory.

yangmillstheory:
Just have one last question: https://github.com/acdlite/redux-actions/pull/168#discussion_r88968660

timche:
Already added :P

yangmillstheory:
Was this test failing before the application code changed?

timche:
What do you mean exactly?

yangmillstheory:
If we removed the FSA check removal, then the test should fail (TDD) to eliminate false positives. Just wondering if that's the case

timche:
The test I've wrote was failing when I've just removed the checks, because `type` was missing.

yangmillstheory:
Great, thanks :)

PR #175, opened 2016-12-09
Add ES-modules build

goto-bus-stop:
Hi :wave: Thanks for y'all's work on redux-actions, it's very useful :)

I do have a small proposal: to publish an ES-modules build as well, for bundlers like Rollup and Webpack2.

This patch conditionally enables the `transform-es2015-modules-commonjs` transform based on an env var. All the environment setup in the npm scripts is getting a little messy this way, though, because the transform needs to be enabled in all cases except one. I don't think there's a way to conditionally _disable_ a transform in a .babelrc unfortunately :(

---

What I could do instead is add a "custom" preset that determines whether to set the `modules` property based on an env var:

```js
// build/preset.js
const esModules = process.env.BABEL_ENV === 'es'
module.exports = {
  presets: [
    ['es2015', { modules: esModules ? false : 'commonjs' }]
  ]
}
```
```js
// .babelrc
{
  "presets": [
    "./build/preset",
    "stage-0"
  ]
}
```

That would keep the npm scripts more clean. If that sounds better, let me know!


yangmillstheory:
👋 

Thanks. I'm out today, but can take a look tonight 👍.

yangmillstheory:
Wait, could you illustrate what problem this is trying to solve? 

I use ES2015 modules with Webpack, so I don't think I've run into the use case.

goto-bus-stop:
It already works fine with Rollup and Webpack2, but they can't apply tree-shaking to the CommonJS module. Also, there's duplicate module interop code: Babel's `requireInterop` code, and then Rollup or another bundler's glue code to make CommonJS work with their preferred format.

With this PR, bundlers that support ES modules can tree-shake: if Rollup builds a project that is eg. not using the `createActions` method anywhere, it won't be added to the bundle (and neither will the lodash methods that `createActions` depends on). And module bundlers that don't support ES modules natively will just keep using the CommonJS build.

yangmillstheory:
Can we try the `build/preset.js` approach, since it seems like it'd be cleaner?

goto-bus-stop:
Sure! Added :)

package.json
```diff
@@ -3,20 +3,24 @@
   "version": "1.1.0",
   "description": "Flux Standard Action utlities for Redux",
   "main": "lib/index.js",
+  "module": "es/index.js",
```
yangmillstheory: What is this `module` key for? Couldn't find anything about it.

build/preset-es2015.js
```diff
@@ -0,0 +1,6 @@
+const esModules = process.env.BABEL_ENV === 'es';
```
yangmillstheory: It looks like `commonjs` is the default, so how about 

```js
const esOptions = process.env.BABEL_ENV === 'es' ? { modules: false } : {}
```

That way we don't have to hardcode `commonjs` and we can just use Babel defaults.

package.json
```diff
@@ -3,20 +3,24 @@
...
+    "build:es": "cross-env BABEL_ENV=es babel src --out-dir es --ignore *-test.js",
     "build:commonjs": "babel src --out-dir lib --ignore *-test.js",
     "build:umd": "cross-env NODE_ENV=development webpack",
     "build:umd:min": "cross-env NODE_ENV=production webpack",
-    "build": "npm run clean && npm run build:commonjs && npm run build:umd && npm run build:umd:min",
-    "clean": "rimraf lib",
-    "lint": "esw src webpack.config --color",
+    "build": "npm run clean && npm run build:es && npm run build:commonjs && npm run build:umd && npm run build:umd:min",
+    "clean": "rimraf lib es",
+    "lint": "esw build src webpack.config --color",
```
yangmillstheory: Why are we linting `build/`? Is there even a `build/` directory?

yangmillstheory:
I wonder if it's worth including a comment (or modifying the existing) in [this section](https://github.com/acdlite/redux-actions/blob/master/README.md#installation) of the README on the ES2015 build. Thoughts?

package.json
goto-bus-stop: It's for the new `build/preset-es2015.js` file.

package.json
goto-bus-stop: `module` serves about the same purpose as `jsnext:main` and is used by [webpack2](https://webpack.js.org/guides/author-libraries/#final-steps) and [Rollup](https://github.com/rollup/rollup-plugin-node-resolve#readme). I went by [this Rollup wiki page](https://github.com/rollup/rollup/wiki/jsnext:main) which says

> If in doubt, use both!

yangmillstheory:
Thanks @goto-bus-stop. I'll publish `v1.2.0` shortly.

yangmillstheory:
https://github.com/acdlite/redux-actions/releases/tag/v1.2.0 has been published.

Issue #197, opened 2017-03-01
handleActions returning defaultState even when state passed is not undefined


amit1911:
```
import { handleActions } from 'redux-actions';

const getInitialState = () => {
    return ['red'];
};

const reducer = handleActions({
    test: (state, action) => {
        return action.payload.a ? undefined : ['yellow'];
    },
    test2: (state) => {
        return ['1'];
    }
}, getInitialState());

const reducer2 = handleActions({
    test: (state, action) => {
        return action.payload.a ? undefined : ['yellow'];
    }
}, getInitialState());

console.log(reducer(['green'], {
    type: 'test',
    payload: {
        a: 1
    }
}));
// ['red']

console.log(reducer2(['green'], {
    type: 'test',
    payload: {
        a: 1
    }
}));
// undefined
```
The first console statement returns the defaultState which is `['red']` whereas the second console statement returns `undefined` even though they both respond to the same action and have same action handling logic.  Can someone tell me why is this happening?

yangmillstheory:
I can't reproduce this. Here's the (full) code snippet:

```js
const { handleActions } = require('redux-actions')

const getInitialState = () => {
    return ['red'];
};

const reducer = handleActions({
    test: (state, action) => {
        return action.payload.a ? undefined : ['yellow'];
    },
    test2: (state) => {
        return ['1'];
    }
}, getInitialState());

const reducer2 = handleActions({
    test: (state, action) => {
        return action.payload.a ? undefined : ['yellow'];
    }
}, getInitialState());

console.log(reducer(['green'], {
    type: 'test',
    payload: {
        a: 1
    }
}));
// ['red']

console.log(reducer2(['green'], {
    type: 'test',
    payload: {
        a: 1
    }
}));
```

And though I don't think it matters:

```
> node -v
v7.4.0
```

amit1911:
That's weird. I can easily reproduce this on http://requirebin.com/

yangmillstheory:
```
example node test.js
undefined
undefined
```

Expected in both cases since `action.payload.a` is truthy (although I'm not sure why you're returning undefined; state shouldn't be undefined).

If I changed the `undefined` return value on requirebin, I get correct output:

![image](https://cloud.githubusercontent.com/assets/2729079/23447294/d2b22854-fdfe-11e6-9653-5a11c2fb0c73.png)

I can repro this on requirebin. Bug on their end?

amit1911:
[...]

PR #203, opened 2017-03-20
Still return defaultState in handleActions reducer when no handlers are provided

### LINKED ISSUE #202: Create store failed when use handleActions with empty reducerMap
I have a `todosReducer`:
```
import { handleActions } from 'redux-actions';

const initState = {}

export default handleActions({

}, initState)
```

Error message:
```
Uncaught Error: Reducer "todos" returned undefined during initialization. If the state passed to the reducer is undefined, you must explicitly return the initial state. The initial state may not be undefined.
```

webholics:
Fixes #202

### DIFF
--- a/src/__tests__/handleActions-test.js
+++ b/src/__tests__/handleActions-test.js
@@ -257,7 +257,7 @@ describe('handleActions', () => {
     wrongTypeHandlers.forEach(wrongTypeHandler => {
       expect(
         () => handleActions(wrongTypeHandler, defaultState)
-      ).to.throw(Error, 'Expected handlers to be an object with reducers');
+      ).to.throw(Error, 'Expected handlers to be an plain object.');
     });
   });
 });
--- a/src/__tests__/handleActions-test.js
+++ b/src/__tests__/handleActions-test.js
@@ -237,8 +237,9 @@ describe('handleActions', () => {
     });
   });
 
-  it('should still return defaultState when no handlers where provided', () => {
+  it('should always return defaultState when no handlers where provided', () => {
     const reducer = handleActions({}, defaultState);
+    expect(reducer()).to.deep.equal({ counter: 0 });
     expect(reducer()).to.deep.equal({ counter: 0 });
   });
 });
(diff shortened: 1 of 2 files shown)


yangmillstheory:
I think I misread https://github.com/acdlite/redux-actions/issues/202#issuecomment-287732099.

Why do you think an empty object is not an usage error? I'd have thought that we would be throwing an error (using `invariant`) and writing a test for that instead.

webholics:
First: It was possible in previous versions of redux-actions. So it was a breaking change for us.
For sure it is an edge case.
But my argument would be that it is still semantically correct to give the `handleActions` function an empty object which would mean "please do not handle any actions (at the moment)". Returning the default state doesn't have anything to do with how many actions the reducer handles (0-infinity) and should always work.

yangmillstheory:
Thanks for the reply. We're on 2.0 now, that breaking change should have been noted. Though I'm not sure when this break occurred.

I can't think of a use case in real apps. Could you give an example?

webholics:
You're right there is no "real" use case for that. However we often start new redux/react apps with an empty skeleton of different reducers and this breaks our skeletons now.
Its not a huge deal. I would also be fine with an invariant check, but it shouldn't just silently break as it does now.

yangmillstheory:
I think defaulting to `defaultState` is still correct for the created reducer. I'm wondering if this would also fix I think this will also fix https://github.com/acdlite/redux-actions/issues/197.

Can we do an invariant check in `handleActions` (maybe using `isPlainObject`), write some tests for the new API and `defaultState` behavior (with a valid action), and we can get this merged and released? Thanks for your time.

webholics:
I update the PR with more tests and the invariant check. Hope the expected behaviour is better described now.

Regarding #197 
It is not fixed yet with this PR. I think the origin of this issue is that the defaultState should not be forwarded to the `handleAction` calls at all. `handleAction` should be called with an undefined defaultState in this case. Let the `handleActions` handle setting the default itself as it is now implemented here. This of course would require the removal of the invariant check of defaultState in `handleAction`. Which for me would be OK because it is totally valid to have an undefined defaultState.

yangmillstheory:
Thanks for the contribution!

PR #206, opened 2017-04-04
Utilize lodash-es in es builds.

jaridmargolin:
This pattern exists in a few libraries that often coexist with redux actions:

https://github.com/reactjs/redux/blob/master/build/use-lodash-es.js
https://github.com/rt2zz/redux-persist/blob/master/tools/use-lodash-es.js

Without this patch, my bundle contains lodash-es and lodash modules. With the patch included, I was able to shave off a few gzipped KBs by sharing the already required/included lodash-es modules.

### DIFF
--- a/build/use-lodash-es.js
+++ b/build/use-lodash-es.js
@@ -1,0 +1,10 @@
+module.exports = function useLodashEs() {
+  return {
+    visitor: {
+      ImportDeclaration(path) {
+        const source = path.node.source;
+        source.value = source.value.replace(/^lodash($|\/)/, 'lodash-es$1');
+      }
+    }
+  };
+};
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
   "dependencies": {
     "invariant": "^2.2.1",
     "lodash": "^4.13.1",
+    "lodash-es": "^4.17.4",
     "reduce-reducers": "^0.1.0"
   }
 }
--- a/.babelrc
+++ b/.babelrc
@@ -2,5 +2,12 @@
   "presets": [
     "./build/preset-es2015",
     "stage-0"
-  ]
+  ],
+  "env": {
+    "es": {
+      "plugins": [
+        "./build/use-lodash-es"
+      ]
+    }
+  }
 }


yangmillstheory:
Thanks @jaridmargolin, I'll take a look this weekend. How many bytes are saved when running https://github.com/acdlite/redux-actions/blob/master/package.json#L12?

build/use-lodash-es.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function useLodashEs() {
+  return {
+    visitor: {
+      ImportDeclaration(path) {
+        const source = path.node.source;
```
yangmillstheory: Does this work?

```js
const { node: { source } } = path;
```

Not a big deal, just curious. It's a bit less verbose.

package.json
```diff
@@ -62,6 +62,7 @@
   "dependencies": {
     "invariant": "^2.2.1",
     "lodash": "^4.13.1",
```
yangmillstheory: Do we still need `lodash` then? I don't have familiarity with `lodash-es` and the change in `.babelrc`, maybe you could give some context?

jaridmargolin: Yes. lodash will still be used in the source code. The only piece that changes is the `es` compiled files.These distribution files are rewritten to utilize lodash-es instead of lodash.

build/use-lodash-es.js
jaridmargolin: I'm not sure. Its a fairly small piece of functionality isolated from the majority of the code base. I personally wouldn't spend too much time on worrying about aesthetic changes.

But if you have a really strong opinion here, I can attempt to make the change.

jaridmargolin:
I don't believe there will be any effect on UMD builds. The purpose here is just to reduce the duplicated lodash module included while building es source.

build/use-lodash-es.js
yangmillstheory: No strong opinion, just thinking out loud.

jaridmargolin: 👍 appreciate you taking a look at everything

jaridmargolin:
🙌 thanks!

yangmillstheory:
@jaridmargolin can you respond to https://github.com/acdlite/redux-actions/issues/208#issuecomment-294919032?

PR #210, opened 2017-05-07
createAction does more pre-processing

Lucretiel:
Slightly simplified createAction, and made it do more processing ahead of time that doesn't have to be done every time an action is emitted

### DIFF
--- a/src/createAction.js
+++ b/src/createAction.js
@@ -23,7 +23,11 @@ export default function createAction(type, payloadCreator = identity, metaCreato
 
   const actionCreator = (...args) => {
     const payload = finalPayloadCreator(...args);
-    const action = { type, error: payload instanceof Error };
+    const action = { type };
+
+    if (payload instanceof Error) {
+      action.error = true;
+    }
 
     if (payload !== undefined) {
       action.payload = payload;
--- a/src/createAction.js
+++ b/src/createAction.js
@@ -9,18 +9,22 @@ export default function createAction(type, payloadCreator = identity, metaCreato
     'Expected payloadCreator to be a function, undefined or null'
   );
 
+  const wrappedPayloadCreator = (head, ...args) => (head instanceof Error 
+    ? head 
+    : payloadCreator(head, ...args));
+
   const finalPayloadCreator = isNull(payloadCreator) || payloadCreator === identity
     ? identity
-    : ((head, ...args) => head instanceof Error ? head : payloadCreator(head, ...args));
-  
+    : wrappedPayloadCreator;
+
   const hasMeta = isFunction(metaCreator);
-  
+
   const typeString = type.toString();
 
   const actionCreator = (...args) => {
     const payload = finalPayloadCreator(...args);
-    const action = {type, error: payload instanceof Error};
-    
+    const action = { type, error: payload instanceof Error };
+
     if (payload !== undefined) {
       action.payload = payload;
     }
--- a/src/createAction.js
+++ b/src/createAction.js
@@ -9,8 +9,8 @@ export default function createAction(type, payloadCreator = identity, metaCreato
     'Expected payloadCreator to be a function, undefined or null'
   );
 
-  const wrappedPayloadCreator = (head, ...args) => (head instanceof Error 
-    ? head 
+  const wrappedPayloadCreator = (head, ...args) => (head instanceof Error
+    ? head
     : payloadCreator(head, ...args));
 
   const finalPayloadCreator = isNull(payloadCreator) || payloadCreator === identity
--- a/src/createAction.js
+++ b/src/createAction.js
@@ -28,7 +28,7 @@ export default function createAction(type, payloadCreator = identity, metaCreato
     if (payload !== undefined) {
       action.payload = payload;
     }
-    
+
     if (hasMeta) {
       action.meta = metaCreator(...args);
     }
--- a/src/createAction.js
+++ b/src/createAction.js
@@ -7,18 +7,19 @@ export default function createAction(type, payloadCreator = identity, metaCreato
   invariant(
     isFunction(payloadCreator) || isNull(payloadCreator),
     'Expected payloadCreator to be a function, undefined or null'
-  );
+  )
 
-  const finalPayloadCreator = payloadCreator === null || payloadCreator === identity
+  const finalPayloadCreator = isNull(payloadCreator) || payloadCreator === identity
     ? identity
-    : (head, ...args) => head instanceof Error ?
-        head : payloadCreator(head, ...args);
+    : ((head, ...args) => head instanceof Error ? head : payloadCreator(head, ...args));
   
   const hasMeta = isFunction(metaCreator);
+  
+  const typeString = type.toString();
 
   const actionCreator = (...args) => {
     const payload = finalPayloadCreator(...args);
-    const action = { type, error: payload instanceof Error };
+    const action = {type, error: payload instanceof Error};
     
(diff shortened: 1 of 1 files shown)


src/createAction.js
```diff
@@ -10,35 +9,30 @@ export default function createAction(type, payloadCreator = identity, metaCreato
     'Expected payloadCreator to be a function, undefined or null'
   );
 
-  const finalPayloadCreator = isNull(payloadCreator)
+  const finalPayloadCreator = payloadCreator === null || payloadCreator === identity
```
yangmillstheory: Not sure why we're not using `isNull` anymore? Also why is the `identity` check needed?

src/createAction.js
```diff
@@ -10,35 +9,30 @@ export default function createAction(type, payloadCreator = identity, metaCreato
...
+    
+    if (hasMeta) {
       action.meta = metaCreator(...args);
     }
 
     return action;
   };
 
-  actionCreator.toString = () => type.toString();
+  const typeStr = type.toString();
```
yangmillstheory: Is the local declaration needed, given that it's used in one place?

src/createAction.js
```diff
@@ -10,35 +9,30 @@ export default function createAction(type, payloadCreator = identity, metaCreato
...
-    const hasError = args[0] instanceof Error;
-
-    const action = {
-      type
-    };
-
-    const payload = hasError ? args[0] : finalPayloadCreator(...args);
-    if (!isUndefined(payload)) {
+    const payload = finalPayloadCreator(...args);
+    const action = { type, error: payload instanceof Error };
```
yangmillstheory: This makes the key `error` always be present. Do we really want this new behavior?

src/createAction.js
```diff
@@ -10,35 +9,30 @@ export default function createAction(type, payloadCreator = identity, metaCreato
...
+    
+    if (payload !== undefined) {
```
yangmillstheory: Wait what's the objection to using `isUndefined`? Does it cost that much bytes that the readability isn't worth it? It's not much less tokens to parse, but still.

src/createAction.js
```diff
@@ -10,35 +9,30 @@ export default function createAction(type, payloadCreator = identity, metaCreato
...
     }
-
-    if (hasError || payload instanceof Error) {
-      // Handle FSA errors where the payload is an Error object. Set error.
-      action.error = true;
-    }
-
-    if (isFunction(metaCreator)) {
+    
+    if (hasMeta) {
```
yangmillstheory: So this is where the savings is?

src/createAction.js
Lucretiel: 1. No reason. I just figured that the purpose of `isNull` is to be a predicate to functions like `filter`, so there's no reason to have the function call overhead here.
2. The `identity` check means that if the `payloadCreator` is `identity`, we can skip the `instanceof Error` check every time it's called, and don't have to use the latter version of the `finalPayloadCreator`.

src/createAction.js
Lucretiel: I don't have a strong preference either way. It doesn't change anything for FSA-compliant applications, and I liked the simplicity of this version. I would have put `meta` and `payload` inline as well, but there was no good way to do that given the limitations of javascript syntax.

src/createAction.js
Lucretiel: It's not a readability thing; just avoiding the function call overhead. I figured that the purpose of `isUndefined` is to be used as a predicate, like in `filter`, but there's no reason to use it here, when the comparison is so simple. Compare to `isFunction`, which has a much more complicated definition.

src/createAction.js
Lucretiel: It means that we don't have to call `type.toString` every time `actionCreator.toString` is called.

src/createAction.js
Lucretiel: Yes, especially given that `isFunction` is surprisingly involved.

Lucretiel:
Obviously, all of this stuff is based on my personal preference. Originally I was only planning on doing the `hasMeta` change, but then I started editing other stuff while I was here. I've given my own justification for all these things, but feel free to give me a firm yes or no on any of them.

yangmillstheory:
So, these are actually quite good reasons. The build is failing because of lint; can you get that to pass and I'll get this merged Thanks.

src/createAction.js
yangmillstheory: 1. re: overhead, this is called just once during when defining the action creator, so do we need to worry about it? If not, I would advise removing the `isFunction` call above and below as well for consistency, but would prefer to just keep as is.

src/createAction.js
yangmillstheory: Makes sense here, sense this is being called dynamically at runtime.

src/createAction.js
yangmillstheory: Can we move this declaration up a scope?

yangmillstheory:
@Lucretiel I fixed some lint issues in the PR. I would like to get this in, but some tests are failing. Could you fix?

Lucretiel:
Yeah, mostly it has to do with the having `error` always be present. I'll just remove that aspect, since I didn't feel that strongly about it anyway.

Lucretiel:
Do you mind if I do a force push? Or did you make changes you feel strongly about?

yangmillstheory:
Looks good now. Can you take a look and tell me if this is OK with you? I'll merge and publish a patch version.

Lucretiel:
Made one final change. Everything should be good now.

yangmillstheory:
Thanks 👍 .

yangmillstheory:
https://github.com/acdlite/redux-actions/releases/tag/v2.0.3

```
~/c/p/redux-actions (45c3a78)|master✓
$ npm search redux-actions
NAME                      | DESCRIPTION          | AUTHOR          | DATE       | VERSION  | KEYWORDS
redux-actions             | Flux Standard…       | =acdlite…       | 2017-05-10 | 2.0.3    |
redux-actions-assertions- | Javascript…          | =dmitry-zaets   | 2016-09-12 | 1.1.0    |
js                        |                      |                 |            |          |
redux-actions-sequences   | Make sequences of…   | =andrevinsky    | 2017-01-17 | 1.1.19   |
npmdoc-redux-actions      | #### api…            | =npmdoc         | 2017-04-24 | 2017.4.… |
redux-actions-helpers     | Small helpers…       | =olegman        | 2016-10-27 | 1.0.2    |
redux-actions-api-addon   | Redux Actions…       | =jkusachi       | 2016-07-14 | 1.1.0    |
redux-actions-utils       | Utility fo Redux…    | =dahel          | 2017-03-27 | 0.0.2    |
xm-redux-actions          | Flux Standard…       | =yeanzhi        | 2017-04-24 | 2.0.2    |
redux-actions-assertions- | Javascript…          | =dmitry-zaets   | 2016-09-12 | 1.1.0    |
js                        |                      |                 |            |          |
redux-actions-sequences   | Make sequences of…   | =andrevinsky    | 2017-01-17 | 1.1.19   |
npmdoc-redux-actions      | #### api…            | =npmdoc         | 2017-04-24 | 2017.4.… |
redux-actions-helpers     | Small helpers…       | =olegman        | 2016-10-27 | 1.0.2    |
redux-actions-api-addon   | Redux Actions…       | =jkusachi       | 2016-07-14 | 1.1.0    |
redux-actions-utils       | Utility fo Redux…    | =dahel          | 2017-03-27 | 0.0.2    |
```

PR #211, opened 2017-05-11
Slight optimizations to handleAction

Lucretiel:
- Removed unnecessary falsey check on action.type
- Optimized the common case where handleAction isn't handling a list of types

The primary impetus for this change is that `redux` checks very aggressively that `action.type` exists, throwing an error otherwise. (See https://github.com/reactjs/redux/blob/e58e207213161042cb5853ddb6acb0d65034c3b0/src/createStore.js#L157-L162).

While I was in there, though, I noticed an easy potential optimization, which is that there's no need to do `_.includes` if there's only one action type being handled.

### DIFF
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -7,8 +7,14 @@ import includes from 'lodash/includes';
 import invariant from 'invariant';
 import { ACTION_TYPE_DELIMITER } from './combineActions';
 
-export default function handleAction(type, reducer = identity, defaultState) {
-  const types = type.toString().split(ACTION_TYPE_DELIMITER);
+export default function handleAction(typeObj, reducer = identity, defaultState) {
+  const type = typeObj.toString();
+  const types = type.split(ACTION_TYPE_DELIMITER);
+  // Common-case optimization: most handlers are only handling one action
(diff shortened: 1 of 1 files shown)


yangmillstheory:
Does this improve runtime perf in a substantial way?

Lucretiel:
I could measure it, but it would mostly depend on how frequently handleAction is called.

yangmillstheory:
I guess there comes a point when rewriting code for the sake of performance becomes premature optimization, and could introduce bugs and code complexity for immeasurable/unclear gain? I appreciate the contribution, though, just saying.

Lucretiel:
Sure, I understand. Feel free to reject. Generally when browsing stuff on github I just make pull-requests willy-nilly whenever it's something I would change in my own project; I won't be offended if you don't think it's worth it.

yangmillstheory:
Well I haven't actually dug deep yet, maybe it turns out that this is better and clearer?

Lucretiel:
So, the pull request really has two parts. The first part, where I remove the falsey check on action.type, seems like a no-brainer to me; it's less code, and allows handleAction to be inline. I feel much less strongly about optimizing handleAction for a single action.

PR #214, opened 2017-05-23
Make default payload creator ignores synthetic events.

knpwrs:
When combining this library with `react-redux` I noticed a whole bunch of errors in my console when firing lots of events:

image

The component causing the errors is written as such:

```jsx
const Counter = ({ increment, decrement, counter }) => (
  <div>
    <h1>Counter: {counter}</h1>
    <button onClick={increment}>+1</button>
    <button onClick={decrement}>-1</button>
  </div>
);
```

One solution is to rewrite the component so the synthetic event from React is not passed to the action creator:

```jsx
const Counter = ({ increment, decrement, counter }) => (
  <div>
    <h1>Counter: {counter}</h1>
    <button onClick={() => increment()}>+1</button>
    <button onClick={() => decrement()}>-1</button>
  </div>
);
```

This has the potential to violate some [linting rules](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md). That's where using something like `withHandlers` from `recompose` comes in:

```jsx
const Counter = ({ increment, decrement, counter }) => (
  <div>
    <h1>Counter: {counter}</h1>
    <button onClick={increment}>+1</button>
    <button onClick={decrement}>-1</button>
  </div>
);

// ...

export default compose(
  pure,
  withHandlers({
    increment: props => () => props.increment(),
    decrement: props => () => props.decrement(),
  }),
)(Counter);
```

But this feels like overkill for what the component is and what it does. This change simply makes the default payload creator return a `null` payload if it is passed a synthetic event, thus enabling the very first example to be valid.

Thoughts?

### DIFF
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -23,6 +23,18 @@ describe('createAction()', () => {
       });
     });
 
+    it('returns a null payload by default when passed a synthetic event', () => {
+      const actionCreator = createAction(type);
+      const syntheticEvent = {
+        persist: () => 'foo'
+      };
+      const action = actionCreator(syntheticEvent);
+      expect(action).to.deep.equal({
+        type,
+        payload: null
+      });
+    });
+
     it('should throw an error if payloadCreator is not a function, undefined, null', () => {
       const wrongTypePayloadCreators = [1, false, 'string', {}, []];
 
--- a/src/createAction.js
+++ b/src/createAction.js
@@ -1,16 +1,22 @@
-import identity from 'lodash/identity';
 import isFunction from 'lodash/isFunction';
 import isNull from 'lodash/isNull';
 import invariant from 'invariant';
 
-export default function createAction(type, payloadCreator = identity, metaCreator) {
+const defaultPayloadCreator = (arg) => {
+  if (arg && typeof arg.persist === 'function') {
+    return null;
+  }
+  return arg;
+};
+
+export default function createAction(type, payloadCreator = defaultPayloadCreator, metaCreator) {
   invariant(
     isFunction(payloadCreator) || isNull(payloadCreator),
     'Expected payloadCreator to be a function, undefined or null'
   );
 
-  const finalPayloadCreator = isNull(payloadCreator) || payloadCreator === identity
-    ? identity
+  const finalPayloadCreator = isNull(payloadCreator) || payloadCreator === defaultPayloadCreator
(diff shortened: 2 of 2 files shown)


BerkeleyTrue:
@knpwrs While this would make my life easier, I don't like the idea of putting code in checking specifically for the react event system.

My approach lately has been to use the function type of mapDispatchToProps to prevent this issue.


```js
function mapDispatchToProps(dispatch) {
  return {
    increment: () => incement(),
    decrement: () => decrement()
  };
}
```

knpwrs:
Yeah, I understand that you don't want to add react-specific code. Thinking I might just export argumentless actions with a little Ramda magic:

```js
import R from 'ramda';
export const increment = R.nAry(0, createAction('INCREMENT'));
export const decrement = R.nAry(0, createAction('DECREMENT'));
export const add = createAction('ADD');
```

Then when it comes time to map I can import all actions and map them together:

```js
import * as actions from '../actions/counter';

const mapDispatchToProps = () => actions;
```

Only the simple (argumentless) actions need to be wrapped in `R.nAry(0)`.

knpwrs:
@BerkeleyTrue another idea just came to mind. Another function, e.g., `createSimpleAction` which just ignores any arguments passed to the payload creator. What do you think of that idea?

While I'm on it, no need for Ramda on the previous example:

```js
export const increment = createSimpleAction('INCREMENT');
export const decrement = createSimpleAction('DECREMENT');
export const add = createAction('ADD');
```

yangmillstheory:
@knpwrs love your tagline, thanks for the PR

knpwrs:
My tagline is pretty awesome isn't it? 😏 

If any of the maintainers are interested in a PR for `createSimpleAction` I can go ahead and make one.

BerkeleyTrue:
This looks like a userland issue more than a library issue.

There are already several examples of how to go about solving this in this thread, but even more the library itself provides a way to do this without the need of additional API.

```js
const click = createAction('CLICK', () => {});
```

With Ramda:

```js
const noop = R.always(undefined);
const click = createAction('CLICK', noop);
```

with Lodash:

```js
const click = createAction('CLICK', _.noop);
```

knpwrs:
Surprised to see Ramda doesn't have a `noop`. I see that they have a few issues tracked about it. I was going to suggest `R.T` as well which works as long as a payload of `true` is acceptable.

```
const click = createAction('CLICK', R.T);
```

PR #227, opened 2017-07-07
Pass shorthand arguments to reducers

monvillalon:
Right now, in order to access the fsa properties you either have to use the the action argument directly or use destructuring:

```js
handleActions({
  ACTION: (state, action ) => {
    return { ...state, data: action.payload.data, user: action.meta.user }
  },
  DESTRUCTURING: (state, { payload: { data }, meta: { user } }) => {
    return { ...state, filters, data, user }
  },
})
```

I'm proposing a third backward compatible form, that simplifies destructuring for the most common cases and I think looks pleasing and simpler, and makes it obvious those should 
be the only properties of an fsa action

```js
handleActions({
  SHORT_HAND_ARGS: (state, action, payload, meta, error ) => {
    return { ...state, data:payload.data, user: meta.user }
  },
  SHORT_HAND_DESTRUCTURING: (state, action, { data }, { meta }, error) => {
    return { ...state, filters, data, user }
  },
})
```

### DIFF
--- a/docs/api/handleAction.md
+++ b/docs/api/handleAction.md
@@ -29,13 +29,24 @@ import { handleAction } from 'redux-actions';
 
 If a `reducer` function is passed, it is used to handle both normal actions and failed actions. (A failed action is analogous to a rejected promise.) You can use this form if you know a certain type of action will never fail, like the increment example above.
 
+The reducer function gets recieves the following arguments
+
+1. `state`: The current redux state
+
+1. `action`: The redux action
+	
+ and as a shorthand, for easier desctructuring:
+3. `payload`: The fsa payload of the action, *action.payload*
+4. `meta`: The metadata of the action, *action.meta*
+5. `error`: Boolean flag determining if this is an error, *action.error*
+
 If the reducer argument (`reducer`) is `undefined`, then the identity function is used.
 
 The third parameter `defaultState` is required, and is used when `undefined` is passed to the reducer.
 
 ###### EXAMPLE
 ```js
-handleAction('APP/COUNTER/INCREMENT', (state, action) => ({
+handleAction('APP/COUNTER/INCREMENT', (state, action, payload, meta, error) => ({
   counter: state.counter + action.payload.amount,
 }), defaultState);
 ```
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -28,6 +28,12 @@ export default function handleAction(type, reducer = identity, defaultState) {
       return state;
     }
 
-    return (action.error === true ? throwReducer : nextReducer)(state, action);
+    return (action.error === true ? throwReducer : nextReducer)(
+      state,
+      action,
+      action.payload,
+      action.meta,
+      action.error === true
+    );
   };
 }
--- a/src/__tests__/handleAction-test.js
+++ b/src/__tests__/handleAction-test.js
@@ -252,4 +252,23 @@ describe('handleAction()', () => {
         .to.deep.equal({ number: 3 });
     });
   });
+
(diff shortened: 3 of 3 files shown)


yangmillstheory:
Great idea 👍, always a fan of reducing (no pun intended) boilerplate.

To clarify, should the `SHORT_HAND_DESTRUCTURING` example be

```js
handleActions({
  SHORT_HAND_ARGS: (state, action, payload, meta, error ) => {
    return { ...state, data:payload.data, user: meta.user }
  },
  SHORT_HAND_DESTRUCTURING: (state, _, { data }, { user }, error) => {
    return { ...state, filters, data, user }
  },
})
```

Also, I wish there were a way to not have the unused argument. Can we introspect using [`length`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length) and act accordingly?

docs/api/handleAction.md
```diff
@@ -29,13 +29,24 @@ import { handleAction } from 'redux-actions';
...
 If a `reducer` function is passed, it is used to handle both normal actions and failed actions. (A failed action is analogous to a rejected promise.) You can use this form if you know a certain type of action will never fail, like the increment example above.
 
+The reducer function gets recieves the following arguments
+
+1. `state`: The current redux state
+
+1. `action`: The redux action
+	
+ and as a shorthand, for easier desctructuring:
+3. `payload`: The fsa payload of the action, *action.payload*
```
yangmillstheory: `s/fsa/FSA`, please

docs/api/handleAction.md
```diff
@@ -29,13 +29,24 @@ import { handleAction } from 'redux-actions';
 
 If a `reducer` function is passed, it is used to handle both normal actions and failed actions. (A failed action is analogous to a rejected promise.) You can use this form if you know a certain type of action will never fail, like the increment example above.
 
+The reducer function gets recieves the following arguments
```
yangmillstheory: `s/recieves/receives`, and we can remove "gets" entirely.

docs/api/handleAction.md
```diff
@@ -29,13 +29,24 @@ import { handleAction } from 'redux-actions';
...
+
+1. `state`: The current redux state
+
+1. `action`: The redux action
```
yangmillstheory: Numbering here is incorrect, should be `2`. Also I would say

> the Flux Standard Action

instead of 

> the redux action

docs/api/handleAction.md
```diff
@@ -29,13 +29,24 @@ import { handleAction } from 'redux-actions';
...
+	
+ and as a shorthand, for easier desctructuring:
```
yangmillstheory: Spelling: "destructuring"

src/handleAction.js
```diff
@@ -28,6 +28,12 @@ export default function handleAction(type, reducer = identity, defaultState) {
       return state;
     }
 
-    return (action.error === true ? throwReducer : nextReducer)(state, action);
+    return (action.error === true ? throwReducer : nextReducer)(
```
yangmillstheory: Can we destructure here:

```js
const { payload, meta, error } = action
const hasError = error === true
return (hasError ? throwReducer : nextReducer)(state, action, payload, meta, hasError)
```

docs/api/handleAction.md
chrisbendel: PS markdown suggests
1. A
1. B
1. C


```
1. A
1. B
1. C
```
as it will auto increment numbers for you, and makes adding a step before/after later on not so tedious reordering all of the numbers

PR #237, opened 2017-08-12
Update Tutorial.md

pensierinmusica:
Fix example for `handleActions()`

### DIFF
--- a/docs/introduction/Tutorial.md
+++ b/docs/introduction/Tutorial.md
@@ -165,11 +165,11 @@ const { increment, decrement } = createActions({
 });
 
 const reducer = handleActions({
-  [increment](state, { payload: { amount } }) {
-    return { counter: state.counter + amount }
+  [increment]: ({ counter }, { payload: { amount } }) {
+    return { counter: counter + amount }
   },
-  [decrement](state, { payload: { amount } }) {
-    return { counter: state.counter + amount }
+  [decrement]: ({ counter }, { payload: { amount } }) {
+    return { counter: counter + amount }
   }
 }, defaultState);
 ```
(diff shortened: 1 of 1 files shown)


timche:
What is the issue that you are fixing?

pensierinmusica:
Towards the end of the tutorial there's this code example:

```js
const { increment, decrement } = createActions({
  'INCREMENT': amount => ({ amount: 1 }),
  'DECREMENT': amount => ({ amount: -1 })
});

  [increment](state, { payload: { amount } }) {
    return { counter: state.counter + amount }
  },
  [decrement](state, { payload: { amount } }) {
    return { counter: state.counter + amount }
  }
}, defaultState);
```

Maybe I'm missing something, but the reducer looks odd to me, as [from the docs](https://redux-actions.js.org/docs/api/handleAction.html#handleactionsreducermap-defaultstate) I understood that `handleActions()` takes a map where the keys should be strings or action creators, and the value should be the reducer function for that action type.

But in the tutorial example there's no colon character between the key and the value of the map.

timche:
Now I got your point.

These are "Shorthand method names" and part of ES2015, so nothing wrong here, but not really beginner friendly, if you are new to ES2015.

docs/introduction/Tutorial.md
```diff
@@ -188,7 +184,7 @@ const {
 } = window.ReduxActions;
 
 const reducer = handleActions({
-  [combineActions(increment, decrement)](state, { payload: { amount } }) {
+  [combineActions(increment, decrement)]: (state, { payload: { amount } }) {
```
timche: `=>` is missing.

pensierinmusica:
Oh, now I see, it's this thing: http://es6-features.org/#MethodProperties

It looks weird, but now I get it, ahaha

If the code is correct, even if a bit cryptical, shall we keep the docs as they are, what do you think?

timche:
It's okay for me to change it to make more beginner friendly.

pensierinmusica:
Thx!  :)

Issue #246, opened 2017-09-27
Choosing actions for reducer with regexp


Jeewes:
Normally you write something like:
```javascript
[SOME_ACTION_TYPE]: someActionTypeHandler,
```
So it picks a specific action type and handles it with specific reducer.

I'd like to be able to "catch" action types with regexp. For example, I want to handle every action type that ends with `_DONE` with a specific `successReducer`.

Don't see a way to do that with the current API. Should we implement something to enable such usecase?

yangmillstheory:
Have you tried combineReducers? Less permissive than a refex, but more intentional.

Jeewes:
Yes, but I'm looking for a more generic solution..

27.9.2017 23.59 "Victor Alvarez" <notifications@github.com> kirjoitti:

hendrul:
This is an important feature, I need it to create generic handlers that manages all actions with type ending in some name

mslipper:
Why not compose the `redux-actions` reducer? Something like this:

```js
const internalReducer = handleActions({ 
  SOME_ACTION_NAME: (state, { payload: { amount } }) {
    return { ...state, counter: state.counter + amount };
  }
});

export function reducer(state, action) {
  if (yourRegex.test(action.type)) {
    return internalReducer(state, SOME_ACTION_NAME);
  }

  return internalReducer(state, action)
}
```

You could easily extract this into a helper method and use it across your app. I'm reticent to add support for regular expression matching since it introduces quite a bit of complexity. What happens when both a regex and string key match? How do we store references to the passed-in regexes, since JS objects can only support string keys? If we're supporting regexes, why not arbitrary functions?

ivadenis:
@mslipper this should be done in the `middleware` in my opinion.

tonypee:
couldn't it be done in the handleAction, if the key of the map is a regex? 

handleActions({
 /^test/: (state, action) { ... } 
})

this is a valid object, and it just means doing the check for regex type and executing the test, would be very simple, no?

tonypee:
haha... and 2 years later i come back wanting to do the same thing.. still no replies!? Is regex a thing yet?

alexander-heimbuch:
I will close this issue, @mslipper answer seems like a valid solution.

PR #252, opened 2017-11-01
Action handlers could now match action types with regexp

hendrul:


hendrul:
fixes #246

mslipper:
See discussion on #246. Will reopen if we reach consensus on how to handle cases like this.

PR #259, opened 2017-12-15
Handle the case when the type is null

Anenth:
When by mistake the type of the action is null, The execution breaks at the library. 
This used to work before https://github.com/reduxactions/redux-actions/commit/bea186f6b2467d3f706b54649b1dd4d6f53ffc98#diff-52ace21827ea0130a413a22cd1697dc8

### DIFF
--- a/src/__tests__/createAction-test.js
+++ b/src/__tests__/createAction-test.js
@@ -6,6 +6,11 @@ describe('createAction()', () => {
   describe('resulting action creator', () => {
     const type = 'TYPE';
 
+    it('should not throw error if type of null is passed', () => {
+      const action = createAction(null);
+      expect(action()).to.deep.equal({type: null});
+    });
+
     it('returns a valid FSA', () => {
       const actionCreator = createAction(type, b => b);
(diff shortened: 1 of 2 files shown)


mslipper:
I can take a look at this this weekend. I need a little time to familiarize myself with the codebase since I've only just joined as a maintainer.

mslipper:
Under what conditions will `type` be null? Redux requires the type to always be defined, so it's a reasonable assumption to make at the library level.

Anenth:
Before this https://github.com/reduxactions/redux-actions/commit/bea186f6b2467d3f706b54649b1dd4d6f53ffc98#diff-52ace21827ea0130a413a22cd1697dc8 it used to work. 
am not able to upgrade the lib because of this.

timche:
We are aligning with FSA, which says the type must be a string.

https://github.com/redux-utilities/flux-standard-action#type

Anenth:
@timche Thats correct, But if mistake an action is called with null, It should not break the application, It should only log a warning message, because it wont have any side effects

PR #264, opened 2018-01-15
Curried version of createAction

### LINKED ISSUE #257: Variant of createAction with curried payload creator
There's often times we wish that we could partially apply the result of `createAction`. We want to bind some parameters to the action creator before the final action is constructed.

I've written up a rough implementation of how this could work and I'm happy to raise a PR for this if you're interested.

```
function identity(x) {
    return x;
}


function createAction(type, fn = identity) {
    let pos = 0;
    const appliedArgs = [];
    const argLen = fn.length;

	return function fnCaller(...args) {
	    pos += args.length;
	    appliedArgs = [...appliedArgs, ...args];

	    if (appliedArgs.length < argLen) {
	        return fnCaller;
	    }

	    const payload = fn(...appliedArgs);

	    return {
	        type,
	        payload,
	        error: payload instanceof Error
	    }
	}

}

// Example usage

const foo = createAction('GO_DO_IT', function foo(a, b, c){ return a + b + c });

const a = foo('OK');
const b = a('LETs');
const c = b('GO'); // Action object is created here.

console.log(c); // { type: 'GO_DO_IT', payload: 'OKLETsGO', error: false }

const d = foo('GET', 'ON', 'UP');

console.log(d)  // { type: 'GO_DO_IT', payload: 'GETONUP', error: false }
```

JamieDixon:
This PR adds a curried version of `createAction` so that the payload creator can be partially applied.

This is useful when you want to bind certain parameters to the payload creator without having to create a createAction factory

For example (before and after)

```
const updateCartSize = createAction(UPDATE_CART_SIZE, (previousSize, newItemsLength) => previousSize + newItemsLength);

// We might want to bind the `previousSize` as part of `mapDispatchToProps`

const mapDispatchToProps = {
 updateCartSize: updateCartSize(currentCartSize)
}

// To do this we need to change our action creator to close over our `previousSize` value:

const updateCartSize = previousSize => createAction(UPDATE_CART_SIZE, (newItemsLength) => previousSize + newItemsLength);
```

Instead of having to that, this PR introduces a curried version of the `createAction` function that allows partial application of the payload creator parameters:

```
const updateCartSize = createCurriedAction(UPDATE_CART_SIZE, (previousSize, newItemsLength) => previousSize + newItemsLength);
```

which brings us back to our preferred API 

The example above is simplified but in our codebase this would really help us with consistency and our FP approach.

Fixes #257

### DIFF
--- a/src/createCurriedAction.js
+++ b/src/createCurriedAction.js
@@ -1,0 +1,5 @@
+import createAction from './createAction';
+import curryN from 'lodash/fp/curryN';
+
+export default (type, payloadCreator) =>
+    curryN(payloadCreator.length, createAction(type, payloadCreator));


mslipper:
LGTM

PR #268, opened 2018-01-23
Allow JavaScript Map to be used as the reducer map to handleActions

### LINKED ISSUE #267: Allow object passed to handleActions to be a Map as well as Object
Since the keys for `handleActions` can be either string (i.e. `"LOG_IN"`) or functions (the result of `createAction`), `handleActions` should allow the mapping object to be a `Map` since functions can't be the keys of plain objects.

JamieDixon:
**What**

Allow JavaScript `Map` type to be used as the `map` value to `handleActions`

**Why**

`handleActions` allows you to give it a function created with `createAction` as the key to a reducer. JavaScript doesn't allow functions to be object keys (strings only) but `Map` lets us use any value as the keys

## Implementation
- [x] Add support for JavaScript `Map` as the map parameter to `handleActions`

## Documentation
- [x] Add documentation for using a JavaScript `Map` as the map parameter to `handleActions`
- [x] Add documentation for using a function as the key to a JavaScript map for `handleActions`

## Tests
- [x] Add a test to assert that `handleActions` works correctly given a JavaScript `Map`
- [x] Add a test to assert that functions  work as map keys in `handleActions`

Fixes #267

### DIFF
--- a/docs/api/handleAction.md
+++ b/docs/api/handleAction.md
@@ -114,3 +114,30 @@ const reducer = handleActions(
   { counter: 0 }
 );
 ```
+
+You can also use an action function as the key to a reduce function instead of using a string const:
+
+
+```js
+const increment = createAction(INCREMENT);
+const decrement = createAction(DECREMENT);
+
+const reducer = handleActions(
+  new Map([
+    [
+      increment,
+      (state, action) => ({
+        counter: state.counter + action.payload
+      })
+    ],
+
+    [
+      decrement,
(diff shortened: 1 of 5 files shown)


mslipper:
Thanks for this, @JamieDixon!

JamieDixon:
@mslipper Any change we could get a release with this change in it please? 😇

mslipper:
I'll get in touch with the old maintainers today - I don't have access to the NPM repo just yet.

PR #269, opened 2018-01-29
Option to handle async actions

davo11122:
Added support to thunk actions, now function will check if payload is a function it will add getState function to it


mslipper:
I think this is somewhat out-of-scope of this library - have you tried the example at https://redux-actions.js.org/docs/middleware/index.html?

timche:
> I think this is somewhat out-of-scope of this library

+1

Closing this for now.

PR #275, opened 2018-03-12
Migrate project to Webpack 4

Simek:
Migrate project from Webpack 1 to 4. Due to small code base and rather simple configuration there is not a lot of code that has changed. All tests passed successful on my machine.

I've left few comments in the diff preview to explain few changes in detail.

Also project URLs in `package.json` has been updated and `yarn.lock` has been added to `.gitignore` file.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,10 +1,11 @@
 {
   "presets": [
-    "./build/preset-es2015",
+    "env",
     "stage-0"
   ],
   "env": {
     "es": {
+      "presets": [["env", { modules: false }]],
       "plugins": [
         "./build/use-lodash-es"
       ]
--- a/package.json
+++ b/package.json
@@ -37,21 +37,21 @@
   ],
   "author": "Andrew Clark <acdlite@me.com>",
   "bugs": {
-    "url": "https://github.com/acdlite/redux-actions/issues"
+    "url": "https://github.com/redux-utilities/redux-actions/issues"
   },
-  "homepage": "https://github.com/acdlite/redux-actions",
+  "homepage": "https://github.com/redux-utilities/redux-actions",
   "repository": {
     "type": "git",
-    "url": "https://github.com/acdlite/redux-actions.git"
+    "url": "https://github.com/redux-utilities/redux-actions.git"
   },
   "license": "MIT",
   "devDependencies": {
     "babel-cli": "^6.7.7",
     "babel-core": "^6.7.7",
     "babel-eslint": "^6.1.1",
-    "babel-loader": "^6.2.4",
-    "babel-preset-es2015": "^6.6.0",
-    "babel-preset-stage-0": "^6.5.0",
+    "babel-loader": "^7.1.4",
+    "babel-preset-env": "^1.6.1",
+    "babel-preset-stage-0": "^6.6.0",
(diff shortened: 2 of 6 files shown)


.babelrc
```diff
@@ -1,10 +1,11 @@
 {
   "presets": [
-    "./build/preset-es2015",
```
Simek: Use `env` property for preset injection instead of generating expected content in separate file.

package.json
```diff
@@ -63,7 +63,8 @@
     "gitbook-cli": "^2.3.0",
     "mocha": "^2.2.5",
     "rimraf": "^2.5.3",
-    "webpack": "^1.13.1"
+    "webpack": "^4.1.1",
+    "webpack-cli": "^2.0.11"
```
Simek: Webpack CLI has been extracted to individual package.

webpack.config.babel.js
```diff
@@ -4,27 +4,6 @@ import path from 'path';
 const { NODE_ENV } = process.env;
 const production = NODE_ENV === 'production';
 
-const plugins = [
-  new webpack.optimize.OccurenceOrderPlugin(),
```
Simek: OccurrenceOrderPlugin is used in Webpack 4 by default.

webpack.config.babel.js
```diff
@@ -4,27 +4,6 @@ import path from 'path';
...
-  new webpack.DefinePlugin({
-    'process.env.NODE_ENV': JSON.stringify(NODE_ENV)
-  })
-];
-
-if (production) {
-  plugins.push(
-    new webpack.optimize.UglifyJsPlugin({
```
Simek: Code optimization and minification is now based on new `mode` property. UglifyJsPlugin is by default used in `production` builds. Running with specified UglifyOptions and with default Webpack configuration yield same effect - no noticeable change in output files size.

If custom Uglify plugin configuration is needed it can be provided using `optimization.minimizer` configuration option.

timche:
Awesome stuff, thanks! Delegating to @mslipper for review and further steps ❤️

mslipper:
This is awesome! Thanks @Simek. Will review today.

mslipper:
Going to run some local tests as well to make sure nothing breaks.

Simek:
Thank you @timche  and @mslipper for the quick response. Let me know if I can help or clarify something for you.

Small side note. In future it will be nice to use `babel-plugin-lodash` functionality over manually 
managing `lodash` imports. But that's optimization for another time.

timche:
> Going to run some local tests as well to make sure nothing breaks.

Shouldn't we run CI? I think it hasn't been set up since we moved the repo to this organization 😱

Let me see if I can get it back up.

timche:
Travis should be back up. Can you try a force push and check if that works?

Simek:
Is Node 4 officially supported? I've didn't notice that information earlier. Unfortunately Webpack 4 requires Node 6+. I'm asking because on Travis library is builded on Node 4 and 6.

timche:
Should be changed to LTS and latest. Can you include Travis changes in your PR please?

mslipper:
@timche thanks for fixing that :). This looks good to me.

PR #279, opened 2018-04-05
Add missing colon to combineActions example

emilbaekdahl:


docs/api/combineActions.md
```diff
@@ -51,7 +51,7 @@ const { increment, decrement } = createActions({
 });
 
 const reducer = handleActions({
-  [combineActions(increment, decrement)](state, { payload: { amount } }) {
```
timche: This is btw correct: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

docs/api/combineActions.md
```diff
@@ -51,7 +51,7 @@ const { increment, decrement } = createActions({
...
+  [combineActions(increment, decrement)]: (state, { payload: { amount } }) {
```
timche: This is incorrect as `=>` is missing for arrow function declaration.

mslipper:
Ah, you are tight. Reverting.

PR #280, opened 2018-04-07
Feature: Default State Creator

DragonRaider5:
Hi,

I'm in a situation where I need to check the local storage in order to initially populate my state (as it holds device preferences which got auto-detected and/or set by the user before).
Works great!

But now I also need to reset my redux state while the application is running (when the user logs out) - which also works nicely, no problems.
However if you put those two features together, you have a problem... The state will always be rehydrated based on the (old, maybe outdated) local storage data from the first page load.

This is why I decided to allow for default state creator functions in handleAction(s)!

I think that I'm not the only one having such a use case (where the default state must be "up to date" at every reset) - that's why I decided to share this PR.
In addition this feature allows for async data fetching in your `getDefaultState()` by being able to make async calls in it, return your bare bone default state and dispatch an action when the async operation finishes.

Example:
```js
import { handleActions } from 'redux-actions'

import store from './store.js'
import { fetchSomeData } from './service.js'
import { fetchSomeData as fetchSomeDataAction } from './actions.js'

const getDefaultState = () => {
  // perform some time of execution critical and/or async task here
  fetchSomeData()
    .then((data) => store.dispatch(fetchSomeDataAction(data))

  const preferences = JSON.parse(window.localStorage.getItem('preferences'))

  return {
    preferences,
    someData: null
  }
}

export default handleActions({
  [fetchSomeDataAction]: (state, action) => ({
    ...state,
    someData: action.payload
  })
}, getDefaultState)
`
``

### DIFF
--- a/docs/api/handleAction.md
+++ b/docs/api/handleAction.md
@@ -31,7 +31,7 @@ If a `reducer` function is passed, it is used to handle both normal actions and 
 
 If the reducer argument (`reducer`) is `undefined`, then the identity function is used.
 
-The third parameter `defaultState` is required, and is used when `undefined` is passed to the reducer.
+The third parameter `defaultState` is required, and accepts either the default state or a function returning your default state. It is used/called when `undefined` is passed to the reducer.
 
 ###### EXAMPLE
 ```js
@@ -75,7 +75,7 @@ import { handleActions } from 'redux-actions';
 
 #### `handleActions(reducerMap, defaultState)` {#handleactionsreducermap-defaultstate}
 
-The second parameter `defaultState` is required, and is used when `undefined` is passed to the reducer.
+The second parameter `defaultState` is required, and accepts either the default state or a function returning your default state. It is used/called  when `undefined` is passed to the reducer.
 
 (Internally, `handleActions()` works by applying multiple reducers in sequence using [reduce-reducers](https://github.com/acdlite/reduce-reducers).)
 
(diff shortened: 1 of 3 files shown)


timche:
Something like this should be handled in a dedicated action when starting the app in `componentDidMount`.

DragonRaider5:
As I already mentioned, this PR isn't about app startup - for this case you could simply execute the function getDefaultState().

It's about when the state has gotten reset while the app is running - for example after a logout.

timche:
We are aligning with the concept of Redux:

> We'll start by specifying the initial state. Redux will call our reducer with an `undefined` state for the first time.

https://redux.js.org/basics/reducers#handling-actions

Which means the initial state is only being used at start up and should stay like this.

DragonRaider5:
Yeah, but what about user-triggered restarts?

A logout means a completely new startup of the app (with clean data) which is user-triggered, which should thereby result in an `undefined` state.

timche:
That's correct, but still `handleActions` shouldn't do any side-effects.

> It's very important that the reducer stays pure. Things you should never do inside a reducer:
>
> - Mutate its arguments;
> - Perform side effects like API calls and routing transitions;
> - Call non-pure functions, e.g. Date.now() or Math.random().

https://redux.js.org/basics/reducers#handling-actions

Please move any side-effects to the action creator and use something like redux-thunk.

PR #290, opened 2018-04-30
Missing round brackets

sppiotrowski:
I am afraid that those round brackets are necessary to run an example.


timche:
Good catch. Can you please run prettier once to fix this PR? Happy to merge it afterwards.

sppiotrowski:
Thank you for a message, I would love to do it. Unfortunately this is first time here :). Could you be so kind and give me a tip how I should proceed?

timche:
Nice, thanks a lot!

PR #295, opened 2018-05-07
Add sideEffects: false to enable tree shaking

anajavi:
This PR enables webpack 4 to do tree shaking on es-build.

It adds sideEffects: false to package.json, which tells webpack that imports don't have sideEffects. https://webpack.js.org/guides/tree-shaking/


timche:
Cool, thanks 👍

PR #302, opened 2018-06-18
Fix syntax error in Readme.md

cinan:


timche:
It's actually [not an error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions), but seems to confuse lots of people.

PR #305, opened 2018-07-16
fix docs combineActions, add missing `:` and `=>`

Isaddo:


timche:
The syntax is actually correct and is method definition on objects: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

I understand that it's not widely known and can cause confusion. I'm happy to merge it if you could resolve the CI errors @Isaddo.

Isaddo:
@timche  Thanks for your information and understanding 

I agree with what you said, Even if the syntax is correct, it can be confusing, and it used computed key at the same time. ~But It's fine If you just close this PR without merge.~ (you have done it before I leave this comment :p)

PR #309, opened 2018-07-25
Use ES5 compatible camelcase library.

nowells:
Fixes https://github.com/redux-utilities/redux-actions/issues/308

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "format:check": "prettier --list-different \"**/*.{js,md}\"",
     "build:es": "cross-env BABEL_ENV=es babel src --out-dir es",
     "build:commonjs": "cross-env BABEL_ENV=commonjs babel src --out-dir lib",
-    "build:umd": "cross-env NODE_ENV=development webpack && es-check es5 dist/redux-actions.js",
+    "build:umd": "cross-env NODE_ENV=development webpack",
     "build:umd:min": "cross-env NODE_ENV=production webpack && es-check es5 dist/redux-actions.min.js",
     "build": "run-s clean && run-p build:**",
     "clean": "rimraf coverage dist es lib",
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
     "format:check": "prettier --list-different \"**/*.{js,md}\"",
     "build:es": "cross-env BABEL_ENV=es babel src --out-dir es",
     "build:commonjs": "cross-env BABEL_ENV=commonjs babel src --out-dir lib",
-    "build:umd": "cross-env NODE_ENV=development webpack",
-    "build:umd:min": "cross-env NODE_ENV=production webpack",
+    "build:umd": "cross-env NODE_ENV=development webpack && es-check es5 dist/redux-actions.js",
+    "build:umd:min": "cross-env NODE_ENV=production webpack && es-check es5 dist/redux-actions.min.js",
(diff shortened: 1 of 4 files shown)


src/utils/camelCase.js
```diff
@@ -7,5 +7,5 @@ export default type =>
     ? camelCase(type)
     : type
         .split(namespacer)
-        .map(camelCase)
+        .map(part => camelCase(part))
```
nowells: due to `camelCase` accepting multiple arguments, and `map` providing the index as a second argument, and entire array as 3rd.

webpack.config.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
     library: 'ReduxActions',
     libraryTarget: 'umd'
   },
+  devtool: 'sourcemaps',
```
nowells: makes the UMD non-minified bundle not be eval sourcemaps.

yarn.lock
```diff
@@ -61,6 +61,10 @@ acorn@^5.0.0, acorn@^5.3.0, acorn@^5.5.0:
   version "5.5.3"
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-5.5.3.tgz#f473dd47e0277a08e28e9bec5aeeb04751f0b8c9"
 
+acorn@~5.4.0:
```
nowells: I don't usually use `yarn` so I have no clue if I updated this properly (just letting `yarn add` do it's thing)

package.json
```diff
@@ -12,8 +12,8 @@
     "format:check": "prettier --list-different \"**/*.{js,md}\"",
     "build:es": "cross-env BABEL_ENV=es babel src --out-dir es",
     "build:commonjs": "cross-env BABEL_ENV=commonjs babel src --out-dir lib",
-    "build:umd": "cross-env NODE_ENV=development webpack",
-    "build:umd:min": "cross-env NODE_ENV=production webpack",
+    "build:umd": "cross-env NODE_ENV=development webpack && es-check es5 dist/redux-actions.js",
+    "build:umd:min": "cross-env NODE_ENV=production webpack && es-check es5 dist/redux-actions.min.js",
```
nowells: I confirmed that this failed before my change (not it passed on the unminified before due to eval sourcemaps obscuring the actual code)

timche:
Thanks a lot for this PR! Would it be possible if we can remove `es-check` again? Because we'd need to add `devtool` otherwise. We can revisit it afterwards to make UMD builds ES5 compatible. As for now `lodash.camelcase` is safe to use in the browser.

nowells:
> Thanks a lot for this PR! Would it be possible if we can remove es-check again? Because we'd need to add devtool otherwise. We can revisit it afterwards to make UMD builds ES5 compatible. As for now lodash.camelcase is safe to use in the browser.

Want me to just remove devtool and only test against the minified version with es-check? Or just remove es-check entirely (and devtool). Totally up to you. Just wanted to add some automated tests around this for future safety.

timche:
Yeah, just replace the package.

yarn.lock
timche: Everything good.

src/utils/camelCase.js
timche: Sure about this? We had it like this before as well with `camelCase` from `lodash`.

nowells:
Thank you so much @timche!

PR #313, opened 2018-07-26
IE11 does not support Array.prototype.includes

nowells:
https://github.com/redux-utilities/redux-actions/pull/307 broke IE11 due to it not supporting Array.prototype.includes

https://caniuse.com/#feat=array-includes

### DIFF
--- a/src/handleAction.js
+++ b/src/handleAction.js
@@ -26,7 +26,7 @@ export default function handleAction(type, reducer = identity, defaultState) {
 
   return (state = defaultState, action) => {
     const { type: actionType } = action;
-    if (!actionType || !types.includes(toString(actionType))) {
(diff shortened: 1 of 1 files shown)


timche:
Great catch, thanks a lot 🙏

nowells:
Thanks for being so quick to incorporate fixes!!! 🙏

timche:
Released in v2.6.1: https://github.com/redux-utilities/redux-actions/releases/tag/v2.6.1

nowells:
Do you want me to try to get free cross browser testing support setup? https://www.browserstack.com/open-source?ref=pricing setup Would be pretty straight forward to setup karma unit tests on browserstack to test on IE11. Give some confidence in the future.

PR #319, opened 2018-08-17
Remove non-existent link

mesaugat:
* Removed middleware link which did not have anything
* Renamed `Change Log` to `Changelog`

Also, https://github.com/redux-utilities/redux-actions/blob/master/docs/advanced/UsageWithMiddleware.md is not linked anywhere? Should we do something about it?


timche:
Thanks!

PR #324, opened 2018-09-17
Embed Link to Flux Standard Action

jacobmoyle:
Enriches documentation by linking to the [Flux Standard Action](https://github.com/redux-utilities/flux-standard-action#flux-standard-action) README.


timche:
Thanks!

Issue #325, opened 2018-09-20
Got an error  'SCRIPT5009 Map is undefined' in IE10


xfields:
Hi,  since `redux-action` is updated to `2.6.1`, I got an error  "SCRIPT5009 Map is undefined" in IE10.

Error got in file `utils/isMap.js`:

```javascript
export default value => value instanceof Map;
```

Should I add `Map` polyfill to my repo ?

Thanks.

PR #326, opened 2018-09-20
IE10 does not support Map (#325)

xfields:


xfields:
My project got en error ([#325](https://github.com/redux-utilities/redux-actions/issues/325)) on IE10 after  I update `redux-actions` to `2.6.1`.

I don't think it is a good idea to add a polyfill by whom use `redux-actions`, they may just don't need the `Map` class.

timche:
Thanks 👍

PR #332, opened 2018-10-16
replace lodash with just, replace webpack with rollup, transpile dependencies

### LINKED ISSUE #331: Importing lodash accounts for more than 50% of this package
https://bundlephobia.com/result?p=redux-actions@2.6.1

screen shot 2018-10-15 at 2 38 23 pm

- self 41%
- lodash.curry 35%
- lodash.camelcase 20%

There are _much_ smaller libraries with similar enough APIs
- [lodash.curry](https://bundlephobia.com/result?p=lodash.curry@4.1.1) 6.5kB
- [just-curry-it](https://bundlephobia.com/result?p=just-curry-it@3.1.0) 286B (more than 22x smaller)
- [lodash.camelcase](https://bundlephobia.com/result?p=lodash.camelcase@4.3.0) 4.6kB
- [just-camel-case](https://bundlephobia.com/result?p=just-camel-case@3.1.0) 401B (more than 11x smaller)

I put together a quick comparison: https://repl.it/@heygrady/camelcase

There was an older issue (https://github.com/redux-utilities/redux-actions/issues/308) about some newer versions of dependencies not working in older browsers. However, that is an issue of bundling. If this package utilized something like rollup to bundle dependencies it wouldn't matter how the imported dependency was written.

It should be possible to make this package around 50% smaller by dropping these weighty dependencies and borrowing [the bundling solution that the main redux library uses](https://github.com/reduxjs/redux/blob/master/rollup.config.js).

heygrady:
fixes #331

- [x] naively copy over the rollup config from redux for umd builds
- [x] preserve babel build for commonjs and esm
- [x] allow babel to parse node_modules (to ensure transpilation of dependencies)
- [x] replace lodash.curry with [just-curry-it](https://github.com/angus-c/just/)
- [x] replace lodash.camelcase with to-camel-case
- [x] upgraded to latest reduce-reducers (increased size)
- [x] upgraded every package to latest

| filename | before | after |
| -------- | ------- | ----- |
| dist/redux-actions.js | 110 KB | 19KB |
| dist/redux-actions.min.js | 20KB (7KB gzipped) | 7KB (3KB gzipped) |

**Notes:**
- Upgrading to latest xo meant disabling a linting rule. I left a note in the code explaining why.


test/camelCase.test.js
```diff
@@ -9,5 +9,5 @@ test('includes forward slashes in words', () => {
 });
 
 test('does nothing to an already camel-cased action type', () => {
-  expect(camelCase('myAction')).toBe('myAction');
+  expect(camelCase('myAction')).toBe('myaction');
```
timche: Please fix this behaviour, because it shouldn't do anything to the string according to the spec.

heygrady:
Wondering if I should revert to the way redux-actions handled the [commonjs and es builds](https://github.com/redux-utilities/redux-actions/blob/master/package.json#L13-L14) before. Using rollup creates a single-file commonjs build, which has benefits but maybe be a jarring API change for users that want to import a file directly from `lib`.

bboydflo:
It would be nice to merge this! Great work @heygrady

timche:
> Wondering if I should revert to the way redux-actions handled the [commonjs and es builds](https://github.com/redux-utilities/redux-actions/blob/master/package.json#L13-L14) before. Using rollup creates a single-file commonjs build, which has benefits but maybe be a jarring API change for users that want to import a file directly from `lib`.

So do u think there any breaking changes?

heygrady:
No. But the previous build did separate files in the lib folder, like this:
https://unpkg.com/redux-actions@2.6.1/lib/

This build will have a single file.

With the new build you can still do “import { handleActions } from ‘redux-actions’”. But you can’t do “import camelCase from ‘redux-actions/lib/utils/camelCase’” like you could before.

timche:
Can we change this to the old behaviour? Because I see this as potentially breaking.

timche:
Thanks 👍

PR #334, opened 2018-11-20
Bugfix: Prevent double-prefixing when using `combineActions`

SuaYoo:
Using the `prefix` option in conjunction with `combineActions` will prefix the combined action type again.  For example:
```js
const options = { prefix: 'my-prefix' };
const { action1, action2 } = createActions('ACTION_1', 'ACTION_2', options)

const combinedActionType = combineActions(action1, action2).toString()

console.log(combinedActionType)
// "my-prefix/my-prefix/ACTION_1||my-prefix/ACTION_2"
```

This PR fixes the issue by checking if the action type string already starts with the prefix + namespace.

### DIFF
--- a/test/handleActions.test.js
+++ b/test/handleActions.test.js
@@ -557,7 +557,7 @@ test('works with combineActions nested', () => {
   });
 });
 
-test.only('works with a prefix and namespace', () => {
+test('works with a prefix and namespace', () => {
   const { increment, decrement } = createActions(
     {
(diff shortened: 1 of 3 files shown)


test/handleActions.test.js
```diff
@@ -556,3 +556,36 @@ test('works with combineActions nested', () => {
     loading: true
   });
 });
+
+test.only('works with a prefix and namespace', () => {
```
timche: Remove `.only`

timche:
Sorry for the long waiting. Thank you very much 🙏 👏

PR #335, opened 2018-11-22
IE11 does not support String.prototype.includes

verebelyin:
#306 broke IE11 due to it's not supporting `String.prototype.includes`

### DIFF
--- a/src/utils/camelCase.js
+++ b/src/utils/camelCase.js
@@ -3,9 +3,9 @@ import camelCase from 'to-camel-case';
 const namespacer = '/';
 
 export default type =>
-  type.includes(namespacer)
-    ? type
+  type.indexOf(namespacer) === -1
+    ? camelCase(type)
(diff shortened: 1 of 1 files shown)


ricale:
It's wrong title. Not `Array.prototype.includes` in this case. `String.prototype.includes` is right.

ricale:
It's wrong title. No `Array.prototype.includes` in this case. `String.prototype.includes` is right.

timche:
Sorry for the long waiting. Thanks for fixing this 👍

vinnymac:
@timche anything I can do to get a `v2.6.5` with this in it? Discovered this broke for one of my clients today running IE11 and I don't want to rollback to `v2.6.3` and break things for IE10 users.

timche:
@vinnymac I'm sorry letting you wait. I've published `v2.6.5`.

PR #340, opened 2019-02-21
Upgrade npm's dependencies to their newest version

szb512:
This command allows npm to upgrade its dependencies to their latest version.

### DIFF
--- a/upgrade.yaml
+++ b/upgrade.yaml
@@ -1,0 +1,2 @@
+Upgrade npm's outdated depencies:
+npm-upgrade


szb512:
Sorry the computer glitched

szb512:
The reason the tests keep failing is because yarn fails to get a file 3 times. It might have stopped the tests.

szb512:
Never mind what I said.

szb512:
I don't think the tests are looking good...

PR #342, opened 2019-03-21
fix(type): fix type detection error

BlackHole1:
1. Previously only judged `null`, but did not judge `undefined`
2. `null == undefined => true`, `null === undefined => false`


BlackHole1:
Ci hangs because it can't find https://registry.yarnpkg.com/event-stream/-/event-stream-3.3.6.tgz

PR #364, opened 2020-06-07
fix xo issues

phapdinh:
This fixes xo errors so when you run npm run lint, you won't see any issues and fixes prettier so your travis will pass

xo_errors


alexander-heimbuch:
Fixed with v3.0.0, still thanks for the PR!

PR #386, opened 2022-12-31
Improve ES build comaptibility 

alexander-heimbuch:
First of all I really love this package and it is one of the must have packages for every redux stack I'm working on <3 Since I'm moving most of my projects to Vite I had some troubles with peer dependencies in this project (invariant) so I made some adjustments to make it more ES friendly and internalised the dependency.

I also streamlined some build tooling and added some exports to the build process.


.travis.yml
```diff
@@ -1,11 +1,17 @@
...
+node_js: 18
+cache:
+  npm: false
+  directories:
+    - "~/.pnpm-store"
+before_install:
+  - curl -f https://get.pnpm.io/v6.16.js | node - add --global pnpm@7
+  - pnpm config set store-dir ~/.pnpm-store
+install:
+  - pnpm install
```
timche, line 11: How about using the official [`pnpm/action-setup`](https://github.com/pnpm/action-setup) instead?

package.json
```diff
@@ -2,28 +2,27 @@
...
+    "dist"
   ],
+  "main": "./dist/redux-actions.umd.js",
+  "module": "./dist/redux-actions.mjs",
+  "exports": {
+    ".": {
+      "import": "./dist/redux-actions.mjs",
+      "require": "./dist/redux-actions.umd.cjs"
+    }
+  },
```
timche, line 25: The question we should ask ourselves actually is if should we continue supporting CommonJS? ESM seems to be what everyone is exporting now and what is becoming the standard in the future to replace CJS. Apparently supporting both can also lead to dual [package hazard](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c?permalink_comment_id=3850849#gistcomment-3850849).

Note: ESM can still import CJS, but not vice versa. If you're already using a modern stack, ESM should be support anyway.

Releasing a new major would also be more justified in this case, making it a pure ESM package.

Related: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

package.json
```diff
@@ -41,74 +40,59 @@
...
-    "prettier": "^1.14.3",
-    "rimraf": "^2.6.2",
-    "rollup": "^0.66.6",
-    "rollup-plugin-babel": "^4.0.3",
-    "rollup-plugin-commonjs": "^9.2.0",
-    "rollup-plugin-node-resolve": "^3.4.0",
-    "rollup-plugin-replace": "^2.1.0",
-    "rollup-plugin-terser": "^3.0.0",
-    "xo": "^0.23.0"
+    "@vitest/coverage-c8": "^0.26.2",
```
timche, line 43: Shouldn't this be the exact version too? I'm assuming pnpm is installing as exact versions.

```suggestion
    "@vitest/coverage-c8": "0.26.2",
```

package.json
```diff
@@ -41,74 +40,59 @@
...
+      "n/file-extension-in-import": "off",
+      "import/extensions": "off",
+      "unicorn/prevent-abbreviations": "off",
+      "unicorn/no-array-for-each": "off",
+      "unicorn/no-for-loop": "off",
+      "padding-line-between-statements": "off",
+      "unicorn/no-array-callback-reference": "off",
+      "n/prefer-global/process": "off",
+      "unicorn/prefer-export-from": "off",
+      "unicorn/prefer-array-flat": "off"
```
timche, line 90: I think for the future we should enable these rules and refactor the code to adhere to the linting rules.

timche (changes requested):
Changes look good to me. Few minor comments.

Other:
- `.DS_Store` should not be included and maybe be ignored with `.gitignore`
- We could reconsider the usage of Codecov and either keep it, replace it or remove code coverage completely

.travis.yml
alexander-heimbuch: I think this is a theoretical problem since I will migrate to Github Actions :)

package.json
alexander-heimbuch: I'm all for a pure ESM package, since this PR doesn't introduce new features we should be good.

.travis.yml
timche: Oops, forgot it Travis, and not GitHub Actions. My bad!

package.json
alexander-heimbuch: Yes totally agree, though some fight with prettier code formatting :D Might need to do some research to sort this out.

timche (approved):
LGTM! Great job! 👏

PR #387, opened 2023-01-22
Add Github Actions

alexander-heimbuch:


.github/workflows/review.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: |
+          echo "STORE_PATH=$(pnpm store path)" >> $GITHUB_OUTPUT
+
+      - uses: actions/cache@v3
+        name: Setup pnpm cache
+        with:
+          path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
```
timche, line 43: While this is good practice, I think for a project of this small size and only 12 dependencies, it's not worth it setting up a whole cache for dependencies. The run itself with cache is about 45 seconds, which is very short. All the extra steps to setting up the cache, downloading it and uploading it may even not make a difference in the end.

I did a quick test without cache and the whole job took only 41 seconds: https://github.com/redux-utilities/redux-actions/actions/runs/4009885719/jobs/6885759017

So in my opinion we could remove this as we're not winning anything here, and even if, we're talking about few seconds, but we're still way under 1 minute, which is fast.

.github/workflows/review.yml
```diff
@@ -0,0 +1,61 @@
+on:
```
timche, line 1: `name` could be added:

```suggestion
name: CI
on:
```

And the filename is still `review.yml` and should be named something like `main.yml` (`master` branch has to be renamed in order for it to make sense)

timche (changes requested):
I've left out the review for `publish.yml`, because it will depend on how `review.yml` will be changed and it should follow it.

.github/workflows/review.yml
```diff
@@ -0,0 +1,61 @@
...
+  pull_request:
+  push:
+    branches: [ master ]
```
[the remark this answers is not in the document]
alexander-heimbuch: it would also run on every branch, so this is by intention

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+        registry-url: 'https://registry.npmjs.org'
```
timche, line 25: I think this is optional and the default value already.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+        version: 7
+
+    - run: pnpm install
+
+    - run: pnpm test
+
+    - uses: bcomnes/npm-bump@v2
+      with:
+        git_email: bcomnes@gmail.com
+        git_username: ${{ github.actor }}
```
timche, line 39: I think we can just associate `github-actions` with this, otherwise this email may be confusing or look suspicious. 

```suggestion
        git_email: github-actions@github.com
        git_username: github-actions
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: pnpm/action-setup@v2
+        id: pnpm-install
+        with:
+          version: 7
+
+      - run: pnpm install
+
+      - run: pnpm format:check
+
+      - run: pnpm format:check
```
timche, line 27: Duplicated command/should be linting instead

```suggestion
      - run: pnpm format:check

      - run: pnpm lint
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+        registry-url: 'https://registry.npmjs.org'
+
+    - uses: pnpm/action-setup@v2
+      id: pnpm-install
+      with:
+        version: 7
+
+    - run: pnpm install
+
+    - run: pnpm test
```
timche, line 34: Should we run `lint` and `format:check` here as well?

timche (changes requested):
Thanks for the changes! Looks good. Few comments on the publish workflow. Otherwise great progress and work! 👍

.github/workflows/publish.yml
alexander-heimbuch: The docs state that it is required for publishing: https://github.com/bcomnes/npm-bump

.github/workflows/publish.yml
alexander-heimbuch: I thought it might block a release, that's why I intentionally left it out.

timche: Wouldn't `lint` still be good to run as it would catch any unused variables, inconsistent code, etc, just in case?

timche (approved):
Think it's looking good now! Just small remark on whether to run `lint` in release or not, but will let you decide.

